PR #29, opened 2025-08-13
Add workflow: githubnext/agentics/issue-triage

Krzysztof-Cieslak:
Automatically created PR to add workflow: githubnext/agentics/issue-triage


pelikhan (approved):

.github/workflows/issue-triage.md
```diff
@@ -0,0 +1,98 @@
...
+    allowed:
+      WebFetch:
+      WebSearch:
+
+# By default agentic workflows use a concurrency setting that
+# allows one run at a time, regardless of branch or issue. This is
+# not appropriate for triage workflows, so here we allow one run
+# per issue at a time.
+concurrency:
+   group: "triage-${{ github.event.issue.number }}"
```
pelikhan, line 28: feels like this kind of "pick the right concurrency" should just work. Our compiler should do the right thing and emit a concurrent id with workflow / issue number

Krzysztof-Cieslak: So this
```
concurrency:
   group: "triage-${{ github.event.issue.number }}"
```
is taken directly from actions synteax.

It feels we should introduce some "higher level" syntax for this thing like `concurrency: per-issue` or `concurrency: single` and let compiler generate action-specific syntax

pelikhan: yes.... concurrency key has no logging output as far as I know . Very hard to trace.

PR #53, opened 2025-08-14
Concurrent download of 10 logs at a time

dsyme:


pelikhan (approved):
So are we gonna hit throttling by doing this? I assume go-gh handles them...

dsyme:
> So are we gonna hit throttling by doing this? I assume go-gh handles them...

More you hit GH API rate limiting eventually. They don't mind if you do 10 or 50 in parallel but you get 5000 an hour

PR #66, opened 2025-08-15
allow time deltas in stop-time

dsyme:
We removed max-runs since it doesn't work

Instead we now allow

```
stop-time: +36h
```

which means "36h after compilation".  There is a known problem that if you change your workflow and recompile the effective stop time will be reset. We figure this is OK for v0.1 where we need some time-limitation mechanism


pelikhan (approved):

pkg/cli/commands.go
```diff
@@ -1139,12 +1145,74 @@ func StatusWorkflows(pattern string, verbose bool) error {
...
 		}
 
-		fmt.Printf("%-30s %-12s %-12s %-10s\n", name, compiled, upToDate, status)
+		fmt.Printf("%-30s %-12s %-12s %-10s %-20s\n", name, compiled, upToDate, status, timeRemaining)
 	}
 
 	return nil
 }
 
+// extractStopTimeFromLockFile extracts the STOP_TIME value from a compiled workflow lock file
```
pelikhan, line 1154: Can we keep commands.go small and move those to the parser package? I will reuse in the logs command

dsyme: we should do a pass to refactor everything into smaller files

dsyme: will do separately

pelikhan: We can ask copilot to do it and add support for this format in the logs command. Should be a one shot.

PR #177, opened 2025-08-21
Only emit task job if needed

dsyme:


pelikhan:
the task job may also have a "if" statement so the generic fix would be to compute the guarding "if" expression and make sure it is empty/true.

pelikhan (approved):

pelikhan:
we need a monster workflow that triggers all these features so that we can flag changes in the generated code in the diff

PR #181, opened 2025-08-21
start to split apart compiler.go a little

dsyme:


pelikhan:
this needs to be coordinate or we'll get into merge hell potentially

pelikhan:
see https://github.com/githubnext/gh-aw/pull/178

PR #184, opened 2025-08-22
output labels

pelikhan:
Support for output that adds labels to current issue/pull_request


docs/frontmatter.md
```diff
@@ -260,6 +260,9 @@ output:
     title-prefix: "[ai] "           # Optional: prefix for PR titles
     labels: [automation, ai-agent]  # Optional: labels to attach to PRs
     draft: true                     # Optional: create as draft PR (defaults to true)
+  labels:
+    allowed: [triage, bug, enhancement] # Mandatory: allowed labels for addition
```
dsyme, line 264: It feels like if this `allowed:` is not specified, then the agentic part should be able to leave the chosen labels as data.

dsyme: This applies to all the `labels:` features.

pelikhan: I'll defer this to another PR.

PR #213, opened 2025-08-25
inline js instead of shared actions

pelikhan:
Always inline JavaScript instead of shared workflows. Creates self contained locked workflows. 

- Logic was moved to JavaScript files + test suite.
- removing old templates


pelikhan:
test proxy fails on forks PRS, no secrets.

pelikhan:
can probably remove more permissions in task/add-reaction. content read not needed anymore.

PR #227, opened 2025-08-26
feat: Add network permissions support for Claude engine

Mossaka:
Signed-off-by: Jiaxiao Zhou <duibao55328@gmail.com>


.github/workflows/example-engine-network-permissions.md
```diff
@@ -0,0 +1,27 @@
...
+
+permissions:
+  contents: read
+
+engine:
+  id: claude
+  permissions:
+    network:
+      allowed:
+        - "docs.github.com"
```
dsyme, line 16: Beautiful! So simple!

.github/workflows/example-engine-network-permissions.lock.yml
```diff
@@ -0,0 +1,557 @@
...
+  secure-web-research-task:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      output: ${{ steps.collect_output.outputs.output }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+      - name: Generate Network Permissions Hook
```
dsyme, line 30: Make sure this is only generated if the feature is used (I haven't checked the code either way)

dsyme:
Very cool work.

pkg/workflow/claude_engine.go
```diff
@@ -69,6 +87,11 @@ func (e *ClaudeEngine) GetExecutionConfig(workflowName string, logFile string, e
 		config.Inputs["model"] = engineConfig.Model
 	}
 
+	// Add settings parameter if network permissions are configured
+	if HasNetworkPermissions(engineConfig) {
+		config.Inputs["settings"] = ".claude/settings.json"
+	}
```
Mossaka: From a security perspective, I'd like to change this to be "deny-by-default", but this would mean that all the existing workflows would potentially break. What do you think, @dsyme ?

.github/workflows/example-engine-network-permissions.lock.yml
Mossaka: yes ofc

dsyme (approved):
Great work!

PR #292, opened 2025-09-03
Check the validity of the JSONL file after it is created

dsyme:
Annoyingly the model wrote invalid content to the JSONL file, horking everything, an invalid control character, likely emoji related

```
{"type": "create-pull-request", "branch": "claude-test-pr", "title": "A Pull Request from Claude", "body": "World\n\n## A Haiku About GitHub Actions and AI\n\nCode flows like rivers,  \nAI agents merge and build  \nAutomation dreams."}
```


pelikhan:
@dsyme there is a port of jsonrepair for go that handles a lot of little idiosyncrasies of llms

https://github.com/kaptinlin/jsonrepair

pelikhan:
There is a Boolean expression tree builder/renderer in expressions.go that we can use to create and/or...

Eventually, I would like to support multiline expressions with comments so that we can "read" those and understand them in the compiled code. Also leads to better diffs

PR #313, opened 2025-09-04
support for custom engine to enable testing without llms

pelikhan:
- custom engine defineable in frontmatter (not documented)
- universal workflows that tests all the safe outputs


dsyme:
Looks good!

PR #465, opened 2025-09-05
Neutral tools

dsyme:
https://github.com/githubnext/gh-aw/issues/298

Simplifying `tools:` down and making them neutral of Claude. 

```
tools:
    web-fetch:    
    web-search:  
    bash:            
```


pelikhan:
I would argue that "fetch" is alway web (javascript `fetch`) and `search` the same.

```
tools:
    fetch:    
    search:  
    bash:      
```

PR #869, opened 2025-09-17
🛡️ Enhance Content Sanitization in Workflow Compiler

dsyme:
## Summary

- Improved content sanitization logic in `collect_ndjson_output.cjs`
- Added more robust handling of URLs, protocols, and command-line flags
- Enhanced safe-outputs workflow instructions in compiler

## Key Changes

- Refined URL sanitization to preserve command-line flags with colons
- Improved protocol redaction while keeping specific patterns intact
- Updated workflow compiler to provide more detailed safe-outputs instructions
- Added comprehensive test cases for various sanitization scenarios

### DIFF
--- a/.github/workflows/ci-doctor.lock.yml
+++ b/.github/workflows/ci-doctor.lock.yml
@@ -73,7 +73,7 @@ jobs:
             main();
       - name: Setup Safe Outputs Collector MCP
         env:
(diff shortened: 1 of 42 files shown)


pelikhan:
We should not allow xml comments which can hide things.

PR #1008, opened 2025-09-25
migrate collect output to type script

pelikhan:
- migrate to typescript
- add inputs validation, sanitization


pelikhan:
@copilot 

- generate a set of strongly typed interface for the known message types in the safe output configuration messages (extends SafeOutputConfig in safe-outputs-config.d.ts)
- apply types in collect_ndjson_output.ts

pelikhan:
@copilot in the collect ndjson output file

- add interface type for a safe job configuration item
- use type when validating those messages 
- use schema defined in "inputs" to validate the fields. Apply sanitize if the type is a string

pelikhan:
@copilot format, recompile

PR #1103, opened 2025-09-30
fix: build script now works on MacOS

Mossaka:
The semantics of `sed -i` behaves different on macOS and Linux. I swtiched to a javascript file for cross-compatibility.


pelikhan (approved):

pelikhan:
You could also add a mac flavor of the build in actions to catch those.

PR #1382, opened 2025-10-08
reorder steps in agent job

pelikhan:
- make sure secret redaction is done first
- removed extra roles checks


pelikhan:
@copilot format, lint ,fix tests

PR #1387, opened 2025-10-09
passing all agent output files

pelikhan:
...


pelikhan:
@copilot merge GITHUB_AW_AGENT_STDIO value into GITHUB_AW_AGENT_OUTPUT_FILES (append stdio after the output files). Then add support for rendering a list of files like

PR #1437, opened 2025-10-10
Update workflow to create discussions instead of issues and adjust related configurations

pelikhan:


pelikhan:
@copilot update create-discussion `category-id` to `category`. match the category field against discussion category ids, then names in the create_discussion.cjs implementation. `category` can be a number or a string.

pelikhan:
@copilot merge main, recompile

PR #1583, opened 2025-10-12
fix http transport

pelikhan:
Fix usage of go-sdk


pelikhan:
@copilot fix tests for the updated implementation.

PR #3069, opened 2025-11-03
Add CLAUDE_CODE_OAUTH_TOKEN support to Claude engine

lucasilverentand:
## Summary

Adds support for `CLAUDE_CODE_OAUTH_TOKEN` as an authentication option for the Claude engine, alongside the existing `ANTHROPIC_API_KEY`.

## Changes

- **Core Engine** (`pkg/workflow/claude_engine.go`): Updated to use `GenerateMultiSecretValidationStep` with both tokens, with `CLAUDE_CODE_OAUTH_TOKEN` taking precedence
- **Tests**: Updated all Claude engine and secret validation tests to verify both token types
- **Documentation**: Updated engine docs, CLI docs, and setup templates with examples for both tokens
- **CLI Support** (`pkg/cli/trial_command.go`): Updated `--use-local-secrets` to check `CLAUDE_CODE_OAUTH_TOKEN` first

## Pattern

Follows the existing pattern from Codex engine (`CODEX_API_KEY || OPENAI_API_KEY`) for consistency:
```yaml
ANTHROPIC_API_KEY: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN || secrets.ANTHROPIC_API_KEY }}
```

## Test plan

- [x] All Claude engine tests pass (33+ tests)
- [x] All secret validation tests pass
- [x] Multi-secret validation works correctly
- [x] Backward compatible with existing `ANTHROPIC_API_KEY` usage


pelikhan (approved):

pelikhan:
@lucasilverentand are you able to test this locally?

lucasilverentand:
> @lucasilverentand are you able to test this locally?

The tests complete locally but I'll verify it locally. I've already caught something I (+claude) did wrong. we should not set it as the same variable rather they should be handled as either CLAUDE_CODE_OAUTH_TOKEN or ANTHROPIC_API_KEY as Claude Code handles them differently internally. After this change I will do another round of local validations. 

Excuse my premature PR, I'll fix it right away.

pelikhan:
No worries. I was wondering whether we needed to mint a token using OpenID

lucasilverentand:
Regarding the OpenID/token minting question: No, we don't need to mint the token. The `CLAUDE_CODE_OAUTH_TOKEN` is meant to be obtained by users externally (likely from Claude Code's authentication system) and configured as a GitHub secret, similar to how `ANTHROPIC_API_KEY` works.

The workflow simply:
1. Validates that at least one secret (`CLAUDE_CODE_OAUTH_TOKEN` or `ANTHROPIC_API_KEY`) is configured
2. Passes both environment variables to the Claude CLI
3. The Claude CLI itself handles the authentication logic and determines which token to use

This is consistent with how other engines work (e.g., Codex with `CODEX_API_KEY`/`OPENAI_API_KEY`) - they all expect pre-configured secrets rather than minting tokens.

lucasilverentand:
@pelikhan As you can see it runs in https://github.com/lucasilverentand/gh-aw-trial/actions/runs/19045577600. What would the next steps be to get this merged?

Let me know if I can do anything! 😄

pelikhan:
It's merged. I will mint a release shortly.

lucasilverentand:
Absolutely amazing, thank you for the high momentum here!

Issue #4167, opened 2025-11-16
make deps-dev fails because of missing prettier/node installation


bmerkle:
small issue when setup a dev environment in a clean ubuntu

`make deps-dev` fails with the error below.

looking into the makefile, it seems that prettier is missing despite it is installed as prerequisite
`npm install -g prettier`

merklbe1@H69LST3:~/gh-aw$ prettier
/mnt/c/Users/merklbe1/AppData/Roaming/npm/prettier: 15: exec: node: not found

possible fix: install node automatically via the dependencies

```
merklbe1@H69LST3:~/gh-aw$ make deps-dev
go mod download
go mod tidy
go install golang.org/x/tools/gopls@latest
go: golang.org/x/tools/gopls@v0.20.0 requires go >= 1.24.2; switching to go1.24.10
go install github.com/rhysd/actionlint/cmd/actionlint@latest
go: github.com/rhysd/actionlint@v1.7.8 requires go >= 1.24.0; switching to go1.24.10
npm install -g prettier

changed 1 package in 1s

1 package is looking for funding
  run `npm fund` for details
Downloading GitHub Actions workflow schema...
Formatting schema with prettier...
make: *** [Makefile:136: download-github-actions-schema] Error 127
```

pelikhan:
Updated dev docs and added check in script.

PR #4170, opened 2025-11-17
Update Makefile to include nodejs and npm installation

bmerkle:
Added npm and nodejs installation to development dependencies. 
possible fix for issue #4167

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,9 +122,16 @@ deps:
 
 # Install development tools (including linter)
 .PHONY: deps-dev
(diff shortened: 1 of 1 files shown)


pelikhan:
not sure how maintained is node in all apt distribution... I think we should check for node but let users figure how to best install it.

Or use the devcontainer

pelikhan:
I have update the instructions and check node... however installing the correct version cross platform can be tricky so that's left to the user.

https://github.com/githubnext/gh-aw/pull/4217

PR #4755, opened 2025-11-25
refactor: simplify dev workflow for testing agent assigment

mnkiefer:
* Switched from using `Octokit` to `getOctokit` for performing GraphQL mutations in both the main and fallback agent assignment logic.
* Added enhanced logging for mutation operations, including token usage and mutation details, to aid debugging and auditability.


mnkiefer:
@copilot Analyze https://github.com/githubnext/gh-aw/actions/runs/19675058618/job/56353957110 and fix the issue in assign-to-agent.cjs. For the graphql call that uses octokit with a new token `GH_AW_AGENT_TOKEN` which has the required permissions. Add more logging if needed.

PR #6220, opened 2025-12-12
chore: consistent campaign naming

mnkiefer:
- Campaign names as `Campaign: <Name>`


mnkiefer:
@copilot Analyse broken checks and suggest fixes.

PR #6226, opened 2025-12-12
feat: add campaign orchestrator

mnkiefer:
Adds a campaign orchestrator flow that turns `.campaign.md` specs into dedicated orchestrator workflows:
- Adds BuildOrchestrator and tests to generate a scheduled & workflow_dispatch orchestrator with campaign context (workflows, memory, metrics, tracker label), a default reporting prompt, and default runs-on/roles.
- Teaches `gh aw compile` to treat `.campaign.md` as specs: validate them, build orchestrators, and compile via CompileWorkflowDataWithValidation.
- Checks in newly compiled orchestrator lockfiles for the initial campaigns (i.e. file diet, incident response, org modernization, security compliance).


mnkiefer:
@copilot Analyze and fix CI / test check failing.

mnkiefer:
@pelikhan:  This reads the campaign specs (`.campaign.md`) and generates a `workflow_dispatch` orchestrator you run like any other workflow.

docs/src/content/docs/guides/campaigns.md
```diff
@@ -272,13 +272,33 @@ gh aw campaign status --json       # JSON status output
...
+Campaign spec files participate in the normal `compile` workflow:
+
+- `gh aw compile` automatically validates all `.github/workflows/*.campaign.md` specs before compiling workflows.
+- Validation checks both the spec itself (IDs, tracker labels, lifecycle state, etc.) and that all referenced `workflows` exist in `.github/workflows/`.
+- If any campaign spec has problems, `compile` fails with a `campaign validation failed` error until issues are fixed.
+
+For experimentation, you can also enable an **orchestrator workflow** per campaign:
+
+```bash wrap
+GH_AW_EXPERIMENTAL_CAMPAIGN_ORCHESTRATOR=1 gh aw compile
```
pelikhan: This should be always on unless the rompt is empty.

pkg/campaign/orchestrator.go
```diff
@@ -0,0 +1,63 @@
...
+
+// BuildOrchestrator constructs a minimal agentic workflow representation for a
+// given CampaignSpec. The resulting WorkflowData is compiled via the standard
+// CompileWorkflowDataWithValidation pipeline, and the orchestratorPath
+// determines the emitted .lock.yml name.
+func BuildOrchestrator(spec *CampaignSpec, campaignFilePath string) (*workflow.WorkflowData, string) {
+	// Derive orchestrator markdown path alongside the campaign spec, using a
+	// distinct suffix to avoid colliding with existing workflows.
+	base := strings.TrimSuffix(campaignFilePath, ".campaign.md")
+	orchestratorPath := base + "-campaign.md"
```
pelikhan: Any particular reason why - instead of . ?

pelikhan (commented):
Are campaigns meant to also be scheduled?

pkg/campaign/orchestrator.go
mnkiefer: This is to avoid collision with the campaign spec file itself (`.campaign.md`).

pelikhan: It is too subtle. Be more explicit, something like 

   .campaign.g.md (generated)

It's a common pattern in .Net/C#

mnkiefer:
> Are campaigns meant to also be scheduled?

No, in my opinion campaigns are metadata and governance definitions that reference workflows, while all actual triggers and schedules live only in regular GitHub agentic workflows. But happy to debate on this, it was just my first intuition. ☺️

pelikhan:
> > Are campaigns meant to also be scheduled?
> 
> 
> 
> No, in my opinion campaigns are metadata and governance definitions that reference workflows, while all actual triggers and schedules live only in regular GitHub agentic workflows. But happy to debate on this, it was just my first intuition. ☺️ 

But you do compile each campaign into agentic workflow and no one is going to go through the pain of running manually workflow dispatch. 

The campaign prompt could be a reporting prompt that runs daily by default. If omitted, we have a base prompt with sensible defaults.

mnkiefer:
@copilot Analyze failing checks.

PR #7347, opened 2025-12-23
Fix Copilot Requests permissions in quick start guide

bewuethr:
The Copilot Requests scope seems to only have "Read" permissions, which are sufficient to get the tutorial to work. The instructions currently say to set to "Read and write", which is confusing because the write permissions don't show up in the UI.


pelikhan (approved):

pelikhan:
Thanks!

PR #8876, opened 2026-01-04
chore: update file reduction campaign project details

mnkiefer:
- Updates the configuration for the file size reduction campaign to reference the new "Project 71" instead.


mnkiefer:
@copilot: Also rename .campaign.md file accordingly and recompile workflow.

PR #10038, opened 2026-01-15
Rename workflow to `Playground: assign-to-agent`

mnkiefer:
- Rename from "Dev" to "Playground: assign-to-agent" to describe what it does.


mnkiefer:
@copilot Recompile this workflow

PR #10216, opened 2026-01-16
Modify campaign generator tools and safe outputs configuration

mnkiefer:
- Updated toolsets configuration and added missing safe outputs.


mnkiefer:
@copilot The campaign generator should just run the Process Safe outputs step, not the standalone noes update project and assign to agent that are run in addition. those should not be standalone in the first place.

PR #10332, opened 2026-01-16
chore: update campaign titles and remove author link styles

mnkiefer:
- Adds titles to two campaign example markdown files
- Removes specific author link styles from the CSS


mnkiefer:
@copilot Fix documentation build fails due to invalid links.

pelikhan (approved):

PR #11073, opened 2026-01-21
🔒 Fix network validation for MCP servers

dsyme:
## Summary

- Enhanced network configuration validation for MCP servers with containers
- Added support for top-level network configuration as an alternative to per-server network config
- Improved error messaging for network security in strict mode

## Details

The changes modify the `validateStrictMCPNetwork` function to:
- Accept a top-level network permissions parameter
- Allow network configuration at either the server or workflow level
- Provide more precise validation for servers with containers
- Maintain backward compatibility with existing network configuration approaches

## Test Coverage

Multiple test cases have been added to verify:
- Server-specific network configurations
- Top-level network configurations
- Mixed configuration scenarios
- Error handling for missing network settings


pelikhan:
Proposed action is to remove completely network specification from mcps currently until we bring it back through the spec and im plement this in the gateway.

dsyme:
The fix in the PR looks like it is using the workflow-level network permissions for validation if there is no MCP-specific network permission present.  Which seems correct.

But yeah a separate cleanup to remove the MCP-specific network permissions could/should be done.

PR #11092, opened 2026-01-21
fix: add field definitions to campaign project updates

mnkiefer:
- Expands project field definitions for campaigns `update-project` safe output step.


mnkiefer:
@copilot Remove invalid/outdated link

PR #11164, opened 2026-01-22
feat: add append-only comments

mnkiefer:
- Introduces `append-only-comments` configuration to allow workflows to create new comments instead of updating existing activation comments (timeline style).
- Enhanced tests to verify behavior when append-only comments are enabled.
- Updated documentation to reflect the new configuration option and its usage.


mnkiefer:
@copilot fix tests
notify_comment_error.cjs(59,46): error TS2339: Property 'appendOnlyComments' does not exist on type

PR #11471, opened 2026-01-23
refactor: update tracker label convention from `z_campain_<id>`

mnkiefer:
- Refactors old label format `campaign:<campaignId>` to `z_campaign_<campaignId>` (to place it at the bottom of labels, not as discoverable/in the way).
- Also updates the campaign discovery logic to simplify GitHub search queries by removing unnecessary parentheses, and adjusts related tests to ensure correct query formatting.


mnkiefer:
@copilot Check if this fixes 0 discovery items we've been seeing.

PR #11902, opened 2026-01-26
chore: update campaign docs

mnkiefer:
- Standardizes the campaign tracking label format across docs,from `campaign:<id>` to `z_campaign_<id>`
- Updates the campaign orchestrator execution model and clarifies the separation of responsibilities between orchestrators and workers.


mnkiefer:
@copilot fix docs build

PR #11962, opened 2026-01-26
Revive disclaimer on docs page

eaftan:


pelikhan:
@eaftan maybe add a xml comment above telling the agent to leave this warning.

PR #12482, opened 2026-01-29
refactor: temporary ID handling

mnkiefer:
- Expands support for temporary IDs, allowing IDs to be 12–64 alphanumeric characters (not just hex).


mnkiefer:
- No longer needed

PR #12782, opened 2026-01-30
chore: rm default copilot assignment for create-issue in campaigns

mnkiefer:
Removed default assignee for create-issue workflow (in campaigns).


mnkiefer:
@copilot fix tests

PR #13006, opened 2026-01-31
Disable daily schedule for Dependabot Burner

mnkiefer:
- Comments out the schedule and skip-if-no-match options.


mnkiefer:
@copilot recompile workflow

PR #13305, opened 2026-02-02
Refactor project URLs and GitHub token in workflow

mnkiefer:
Updated project URLs and GitHub token in smoke-project workflow.


mnkiefer:
@copilot Check that project url is set everywhere in workflow consistently and also token where needed. Also, recompile workflow.

PR #13318, opened 2026-02-02
chore: temp disable scheduled triggers for smoke-project

mnkiefer:
- Commented out scheduled triggers and safe outputs in smoke-project workflow for limited testing until stable.


mnkiefer:
@copilot recompile workflow

PR #14479, opened 2026-02-08
chore: update test pr and issue number in `smoke-project`

mnkiefer:
- Reply on pre-existing issue/pr to add to project:
#14478
#14477


mnkiefer:
@copilot recompile workflow only

PR #15279, opened 2026-02-13
🔧 Fix repository checkout logic in workflow compiler

dsyme:
## Summary

- Simplified and standardized repository checkout logic in workflow compiler
- Always add checkout step unless it's already present in custom steps
- Extracted common action reference generation logic into a separate function

## Key Changes

- Created `getActionRef()` function to centralize action reference generation
- Modified `shouldAddCheckoutStep()` to always add checkout by default
- Updated related tests to reflect new checkout behavior
- Removed complex conditional logic for checkout step determination

### DIFF
--- a/pkg/cli/copilot_setup.go
+++ b/pkg/cli/copilot_setup.go
@@ -13,13 +13,18 @@ import (
 
 var copilotSetupLog = logger.New("cli:copilot_setup")
 
+// getActionRef returns the action reference string based on action mode and version
+func getActionRef(actionMode workflow.ActionMode, version string) string {
+	if actionMode.IsRelease() && version != "" && version != "dev" {
+		return "@" + version
+	}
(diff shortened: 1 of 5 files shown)


pkg/cli/copilot_setup.go
```diff
@@ -13,13 +13,18 @@ import (
 
 var copilotSetupLog = logger.New("cli:copilot_setup")
 
+// getActionRef returns the action reference string based on action mode and version
+func getActionRef(actionMode workflow.ActionMode, version string) string {
+	if actionMode.IsRelease() && version != "" && version != "dev" {
```
pelikhan, line 18: Dev contains "dirty" sometimes

dsyme: This is just factoring out code

PR #15506, opened 2026-02-13
Better handling for PR Approval and make footers optional

strawgate:


pelikhan:
Footers are already options through `footer: false`

pelikhan:
I think your branch is behind. Please generate a full description of the changes. thanks.

PR #15693, opened 2026-02-14
perf: Ensure Docker daemon exists AND is running

strawgate:
Fixes https://github.com/github/gh-aw/issues/15690

When Docker Desktop is installed but not running (common when my battery is low), every docker command hangs waiting for the daemon. The existing validation function would then:

1. Run 'docker image inspect' (hangs until timeout)
2. Run 'docker pull' with 3 retries and exponential backoff (5s + 10s + 20s waits between retries)

This adds up to 35+ seconds per container image. With multiple workflows using containers, make recompile could hang for minutes.

Fix: Run a cached 'docker info' check with a 3-second timeout before any docker operations. If the daemon is not responsive, skip container image validation and print a warning. The check runs once and the result is cached for the process lifetime.

### DIFF
--- a/pkg/workflow/docker_validation.go
+++ b/pkg/workflow/docker_validation.go
@@ -101,9 +101,8 @@ func validateDockerImage(image string, verbose bool) error {
 	// This prevents multi-minute hangs when Docker Desktop is installed but not running,
 	// which is common on macOS development machines.
 	if !isDockerDaemonRunning() {
-		dockerValidationLog.Print("Docker daemon not running, skipping image validation")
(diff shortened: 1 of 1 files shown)


pelikhan:
Should be an error

strawgate:
tough but fair, will update

pelikhan (approved):

pelikhan:
the other code path fails so staying consistent... we can maybe have a --validate-maybe flag in the cli

strawgate:
it looks like maybe we've now got slightly inconsistent behavior between `docker` not existing and `docker` not running where docker not running returns an error but docker not existing just prints a log line?

edit: just saw your comment, taking a look

strawgate:
should be consistent now

PR #16042, opened 2026-02-16
Fix concurrent map panic in MCP server caches

AI-Reviewer-QS:
## Summary

The MCP server's `permissionCache` (map) and `repoCache` (pointer) are package-level variables accessed concurrently from HTTP handler goroutines without any synchronization. In Go, concurrent read+write on a plain map causes a fatal runtime panic (`concurrent map read and map write`), crashing the server.

This is triggered when the MCP server runs in HTTP transport mode (`--port`) and receives multiple simultaneous requests that hit `queryActorRole` or `getRepository`.

**Fix:** Add a `sync.RWMutex` (`cacheMu`) to protect both caches. Read operations use `RLock` for concurrency, write operations use `Lock` for exclusive access.

**Affected code paths:**
- `getRepository()` — reads/writes `repoCache` pointer (lines 65-101)
- `queryActorRole()` — reads/writes/deletes from `permissionCache` map (lines 118-150)

## Test plan
- [x] Added `TestQueryActorRole_ConcurrentCacheAccess` — 20 goroutines × 100 iterations exercising concurrent cache read/write/delete
- [x] Added `TestGetRepository_ConcurrentCacheAccess` — same pattern for the repo cache
- [x] Both tests pass with `go test -race` (no data races detected)
- [x] Verified the race is detected without the fix via `go test -race`

### DIFF
--- a/pkg/cli/mcp_server.go
+++ b/pkg/cli/mcp_server.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/github/gh-aw/pkg/console"
@@ -51,6 +52,7 @@ type repositoryCache struct {
 }
 
 var (
+	cacheMu            sync.RWMutex
 	permissionCache    = make(map[string]*actorPermissionCache)
 	permissionCacheTTL = 1 * time.Hour
 	repoCache          *repositoryCache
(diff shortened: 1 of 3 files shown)


pelikhan (changes requested):
Create a class to hide the details and apply locks consistentely.

pkg/cli/mcp_server.go
```diff
@@ -38,44 +39,102 @@ func mcpErrorData(v any) json.RawMessage {
...
 	timestamp  time.Time
 }
 
-var (
-	permissionCache    = make(map[string]*actorPermissionCache)
-	permissionCacheTTL = 1 * time.Hour
-	repoCache          *repositoryCache
-	repoCacheTTL       = 1 * time.Hour
-)
+func newMCPCacheStore() *mcpCacheStore {
```
pelikhan: Move cache to own file

pelikhan (changes requested):

pelikhan (approved):

PR #16289, opened 2026-02-17
feat: compile gh-aw compiler to WebAssembly for browser usage

Mossaka:
## Summary

- Adds `GOOS=js GOARCH=wasm` build target (`make build-wasm`) that compiles the gh-aw workflow compiler to run entirely in the browser
- Uses Go build tags to swap TUI libraries and os/exec-dependent code with lightweight stubs
- Provides a Promise-based JS API: `compileWorkflow(markdown)` → `{yaml, warnings, error}`
- Includes an interactive playground (`wasm-playground/`) and a live editor docs page (`/reference/live-editor/`)
- Wasm binary (~17MB) is gitignored and generated at deploy time via `scripts/bundle-wasm-docs.sh`

## What's included

| Area | Files | Description |
|------|-------|-------------|
| Build tags | ~20 existing files | `!js && !wasm` constraints on files using lipgloss, bubbletea, huh, os/exec, go-gh |
| Wasm stubs | ~20 new `_wasm.go` files | Plain-text/no-op implementations for browser builds |
| String API | `compiler_string_api.go` | `ParseWorkflowString()` + `CompileToYAML()` for in-memory compilation |
| Entry point | `cmd/gh-aw-wasm/main.go` | Registers `compileWorkflow` global via `syscall/js` |
| JS library | `wasm-playground/lib/`, `docs/public/wasm/` | Browser loader, Web Worker wrapper |
| Playground | `wasm-playground/index.html` | Split-panel editor for testing |
| Docs | `reference/wasm-compilation.md`, `reference/live-editor.mdx` | Reference docs + live editor page |

## Test plan

- [x] `GOOS=js GOARCH=wasm go build` succeeds
- [x] Native `go build ./cmd/gh-aw` still works (no regressions)
- [x] Unit tests pass: `pkg/tty`, `pkg/styles`, `pkg/console`, `pkg/workflow`, `pkg/parser`
- [x] Playwright browser test: wasm loads, compiles sample workflow, produces valid YAML
- [x] `make build-docs` succeeds with new docs pages
- [ ] Manual: open playground in browser, paste workflow markdown, verify YAML output


Mossaka (commented):
## PR Review: feat: compile gh-aw compiler to WebAssembly for browser usage

Overall this is a solid, well-structured PR. The build tag strategy is clean, the API surface is small, and the separation between native and wasm stubs is consistent. The playground UIs are polished. Below are my findings organized by review area.

---

### 1. Correctness: Build Tags

**Build tags are applied consistently.** Every native file gets `//go:build !js && !wasm` and every wasm stub gets `//go:build js || wasm`. Shared files (types, utilities) correctly have no build constraint.

The `RepositoryFeatures` struct is correctly duplicated in both `repository_features_validation.go` (native, `!js && !wasm`) and `repository_features_validation_wasm.go` (wasm, `js || wasm`) since the native file is excluded from the wasm build.

The `console_types.go` extraction pattern (pulling `ErrorPosition`, `CompilerError`, `TableConfig`, `TreeNode`, `SelectOption`, `FormField`, `ListItem`, `ToRelativePath`, `RenderTableAsJSON`, `FormatErrorWithSuggestions`, `renderTreeSimple`, `findWordEnd` into a shared file) is a good approach. No symbols are missing.

**One minor observation**: `pkg/cli/` files (which import lipgloss, bubbletea, huh) do NOT have wasm build tags, but this is fine because the wasm entry point (`cmd/gh-aw-wasm/main.go`) only imports `pkg/workflow` and doesn't transitively pull in `pkg/cli`.

**No files appear to be missing constraints.**

---

### 2. API Design: ParseWorkflowString / CompileToYAML

The API is clean and well-designed:

```go
compiler := workflow.NewCompiler(
    workflow.WithNoEmit(true),
    workflow.WithSkipValidation(true),
)
wd, err := compiler.ParseWorkflowString(markdown, "workflow.md")
yaml, err := compiler.CompileToYAML(wd, "workflow.md")
```

**Strengths:**
- Two-phase (parse then compile) gives callers flexibility to inspect `WorkflowData` between steps
- `contentOverride` field with `defer` cleanup is a clean way to thread content through without changing every downstream signature
- The `virtualPath` parameter is well-documented and useful for error messages

**Minor suggestion**: In `doCompile()` in `cmd/gh-aw-wasm/main.go`, a new `Compiler` is allocated per call. For a playground with rapid auto-compile (400ms debounce), this could create GC pressure in the Go wasm runtime. Consider caching a single `Compiler` instance at the module level:

```go
var sharedCompiler = workflow.NewCompiler(
    workflow.WithNoEmit(true),
    workflow.WithSkipValidation(true),
)
```

This would need to ensure `Compiler` is safe for sequential reuse (it appears to be -- `CompileToYAML` resets `stepOrderTracker`, `scheduleFriendlyFormats`, and `artifactManager`). However, `ParseWorkflowString` sets/clears `contentOverride` with defer, which would not be safe for *concurrent* use. Since Go wasm is single-threaded via the event loop, sequential reuse should be safe, but verify this assumption.

---

### 3. Security

**XSS vulnerability in live-editor.mdx (line ~699-704):**

```javascript
loadingEl.innerHTML =
    '<div class="le-loading-error">' +
    '<p><strong>Failed to load the WebAssembly compiler</strong></p>' +
    '<p>' + (err.message || String(err)) + '</p>' + ...
```

`err.message` is interpolated directly into `innerHTML`. While the error originates from a `fetch()` or `WebAssembly.instantiate()` call (making exploitation unlikely in practice), this should still use `textContent` on a created element for defense-in-depth. The wasm-playground's `showValidation()` already demonstrates the correct pattern using `createElement` + `textContent`.

**SRI hashes on CDN scripts (wasm-playground/index.html):** Good -- `js-yaml` and `ajv` both have `integrity` and `crossorigin` attributes. This was noted as a fix in the third commit.

**The standalone editor (`docs/public/editor/index.html`) correctly uses `textContent` for all user-controlled data** (error messages, YAML output, etc.). The only `innerHTML` usage is for line numbers (safe, as it's just generated `<div>N</div>` from a counter).

**The wasm entry point properly rejects unexpected arguments** (line 26-28 in main.go), preventing potential misuse of an `importResolver` argument that isn't supported.

---

### 4. Missing `WebAssembly.instantiateStreaming` fallback in compiler-worker.js

The `wasm-playground/lib/gh-aw-compiler.js` has a proper fallback:

```javascript
try {
    result = await WebAssembly.instantiateStreaming(fetch(...), go.importObject);
} catch (streamErr) {
    const buf = await fallbackResp.arrayBuffer();
    result = await WebAssembly.instantiate(buf, go.importObject);
}
```

But `docs/public/wasm/compiler-worker.js` (used by the docs site live editor) calls `WebAssembly.instantiateStreaming` without any fallback (line 34). If the server serves `.wasm` without the correct `application/wasm` MIME type, the worker will fail. The worker should have the same fallback pattern as the library.

---

### 5. Cross-platform Makefile issue

```makefile
BEFORE=$$(stat -c%s gh-aw.wasm);
```

`stat -c%s` is GNU/Linux-specific. On macOS, the equivalent is `stat -f%z`. Since this is only in the optional `wasm-opt` section (which is informational), it won't break the build, but it will print errors on macOS when `wasm-opt` is available. Consider using `wc -c < gh-aw.wasm` or `du -b` as a portable alternative, or add a platform check.

---

### 6. Testing

The `compiler_string_api_test.go` is solid with good coverage:
- Basic parsing, missing frontmatter, invalid YAML, shared workflow detection
- Virtual path cleaning, empty content, frontmatter-only
- End-to-end round-trip (parse -> compile -> verify YAML structure)

**Missing test coverage:**
- **No test for concurrent calls to `ParseWorkflowString`**: The `contentOverride` field is set on the struct, making concurrent use unsafe. While this is documented by the sequential nature of wasm, a test (or at least a comment in the API doc) would make this explicit.
- **No test for very large workflows**: Would be useful to verify the wasm build doesn't hit memory limits or performance issues with realistic workflow sizes.
- **The Playwright browser test is listed in the test plan but I don't see it in the diff**: Is it in a separate PR or was it omitted from this diff?

---

### 7. Documentation

The `wasm-compilation.md` is thorough and accurate:
- Build tag convention is clearly explained
- Architecture diagram maps stubs to packages
- Limitations table is comprehensive
- The Go version prerequisite was correctly updated from 1.23 to 1.25

**One inaccuracy**: The doc says "Import resolution (`importResolver` callback) is not currently supported" but then says "Local file-based imports and the JavaScript import resolver callback still work" in the "Remote imports" section. These contradict each other. Based on the code (the wasm entry point rejects a second argument, and `remote_fetch_wasm.go` returns errors for workflow specs), the first statement is correct -- no import resolution works in wasm.

---

### 8. Minor Issues

1. **`progress_shared.go` has no build tag** -- correct, but `formatBytes` is a trivial utility. It might be cleaner in a `_shared.go` naming convention or with an explicit comment explaining why it lacks a build tag. Currently it's the only `*_shared.go` file in the package.

2. **`github_cli_wasm.go` imports `os/exec`**: The comment explains why (return type compatibility), and `os/exec` does compile for `GOOS=js GOARCH=wasm`. However, the stubs call `exec.Command(...)` which allocates a `Cmd` that can never run. Consider returning `nil` with a comment, or if callers nil-check, that could be cleaner. But I understand the constraint -- callers expect `*exec.Cmd`.

3. **Redundant `escapeHtml` function** in `wasm-playground/index.html` (line 696-699): This function is defined but never called anywhere in the file. It can be removed.

---

### Summary

This is a well-executed PR. The build tag strategy is consistent, the API is clean, and the code quality is high. The main actionable items are:

1. **Fix the XSS in `live-editor.mdx`** -- use `textContent` instead of `innerHTML` for the error message
2. **Add `instantiateStreaming` fallback to `compiler-worker.js`** -- match the pattern in `gh-aw-compiler.js`
3. **Fix the contradictory import resolution statement in `wasm-compilation.md`**
4. **Consider the `stat -c%s` portability issue in the Makefile** for macOS developers

Everything else is either minor or informational. Nice work!

PR #16304, opened 2026-02-17
chore: clean up standalone editor HTML (10% reduction)

Mossaka:
## Summary
- Removed wrapper divs, duplicate CSS/JS, unused variables. 1317→1187 lines.

## Test plan
- [ ] Verify the standalone editor loads correctly at `/editor/`
- [ ] Check light and dark theme rendering
- [ ] Confirm compile functionality still works


pelikhan:
@Mossaka move the js code in own file for sanity.

Mossaka:
## Addressed review feedback

**Comment 1 — `resizePanels()` cursor fix:**
The `mousedown` handler now checks `window.innerWidth < 768` and sets `document.body.style.cursor` to `'row-resize'` for mobile (vertical layout) or `'col-resize'` for desktop (horizontal layout).

**Comment 2 — Cache `lineNumbers` element:**
The `lineNumbers` element is now cached once at initialization (alongside all other DOM elements) and `syncLineNumberScroll()` uses the cached reference instead of calling `getElementById('lineNumbers')` on every scroll event.

**Additionally — Extracted inline JS to separate file:**
All JavaScript has been moved from the inline `<script>` block in `index.html` into a separate `editor.js` file, loaded as `<script type="module" src="editor.js"></script>`. Module scripts are deferred by default, so DOM references work correctly.

Verified with Playwright:
- Desktop (1280x800): compilation works, `col-resize` cursor during drag
- Mobile (375x812): vertical layout, `row-resize` cursor during drag
- External `editor.js` loaded as module, zero inline scripts

PR #16372, opened 2026-02-17
Fix nested remote imports to resolve relative to parent's base path

strawgate:
- Nested relative imports from remote workflowspecs now resolve relative to the parent file's directory (standard relative path semantics), instead of hardcoded `.github/workflows/`
- `../` traversals supported via `path.Clean` for sibling-directory imports
- Consolidated all inline workflowspec formatting into `FormatWorkflowSpec` helper
- Integration tests against `githubnext/agentics` verify end-to-end nested import resolution

## Problem

`parseRemoteOrigin` stripped the last 2 path components to derive `BasePath`, so `owner/repo/workflows/code-simplifier.md` got an empty base path. Its nested import `shared/reporting.md` resolved at the repo root instead of `workflows/shared/reporting.md`, causing 404s, security scanner warnings, and broken `runtime-import` paths in `.lock.yml` files.

## Fix

`BasePath` is now the parent directory of the file (strip filename only):

| Workflowspec | BasePath |
|---|---|
| `owner/repo/workflows/code-simplifier.md` | `workflows` |
| `owner/repo/base/subdir/file.md` | `base/subdir` |
| `owner/repo/file.md` | `` (repo root) |

`ResolveNestedImport` joins base path + relative import with `path.Clean`, so `../sibling/file.md` from `base/subdir` resolves to `base/sibling/file.md`. The traversal guard now only rejects paths escaping the repository root.

Fixes #16370


strawgate:
will test this locally for my usecase

pkg/parser/import_processor.go
```diff
@@ -491,21 +514,27 @@ func processImportsFromFrontmatterWithManifestAndSource(frontmatter map[string]a
...
+					var basePrefix string
+					if item.remoteOrigin.BasePath != "" {
+						basePrefix = item.remoteOrigin.BasePath + "/"
+					}
 
-						resolvedPath = fmt.Sprintf("%s/%s/.github/workflows/%s@%s",
-							item.remoteOrigin.Owner, item.remoteOrigin.Repo, cleanPath, item.remoteOrigin.Ref)
-						nestedRemoteOrigin = item.remoteOrigin
-						importLog.Printf("Resolving nested import as remote workflowspec: %s -> %s", nestedFilePath, resolvedPath)
+					resolvedPath = fmt.Sprintf("%s/%s/%s%s@%s",
```
dsyme: This spec-formation logic should be in a helper with a good name, and the codebase checked carefully to see if we already have the logic for generating these specs elsewhere.

dsyme (changes requested):
Looks OK but I'd really like to see any string-hacking around the different kind of specs we use put into helpers and shared across the codebase, preferably paired with their parsing, and under independent unit test

pelikhan:
@strawgate let's add integration tests using some samples of githubnext/agentics

strawgate:
Sounds good, let me get the right behavior working in this PR and then i'll expand with tests and identify helper extractions. 

Looks like the problem also occurs for runtime imports -- they are not pointing at the cache when added from a remote source

strawgate:
fyi githubnext/agentics has a one-level import via the shared folder but no multi-level/nested imports

pelikhan:
> fyi githubnext/agentics has a one-level import via the shared folder but no multi-level/nested imports

what do I need to add in there to trigger this code path?

pelikhan:
@copilot merge main and recompile

strawgate:
The repo currently has

Workflow A -> fragment B

I.e. A workflow which uses include to include a fragment 

Ideally we'd also add:

Workflow A -> fragment B -> fragment C

I.e. fragment B has an include in it as well

Basically two levels of nested includes instead of just one

strawgate:
See https://github.com/githubnext/agentics/pull/170 for an example of the double nested workflow

pelikhan:
@copilot add integration test suite that generates various agentic workflows with multiple levels of imports and self adds the workflows to test those code paths. Test extensive graphs combination with combination of file path depth.

pelikhan:
See https://github.com/github/gh-aw/pull/16550

strawgate:
@pelikhan looks like this can close then!

Copilot stealing my hard work is going to have a big impact on my attempt to goose my contributor ranking. 🤣

pelikhan:
We are closing down 3rd party PR/ in favor of issue plans. ;)

PR #16450, opened 2026-02-17
Skip ASCII art header in wasm/editor compiled output

Mossaka:
## Summary
- Skip the large ASCII art banner (`GenerateWorkflowHeader`) when compiling via `CompileToYAML` (the wasm/editor code path)
- The banner takes up the entire visible area in the live editor, hiding the actual YAML output
- Adds a `skipHeader` field to the `Compiler` struct, set to `true` in `CompileToYAML`, checked in `generateWorkflowHeader`

## Test plan
- [x] `GOOS=js GOARCH=wasm go build ./cmd/gh-aw-wasm` passes
- [x] `go build -o /dev/null ./cmd/gh-aw` passes
- [x] `go test -short ./pkg/workflow/` passes
- [x] Updated `TestCompileToYAML_EndToEnd` to verify header is absent in wasm output


pelikhan:
Add tests

PR #16514, opened 2026-02-18
Add inline-prompt option to compile workflows without runtime-import macros

### LINKED ISSUE #16511: Add `inline-prompt` option to compile workflows without runtime-import macros
Workflows compiled with `gh aw compile` emit `{{#runtime-import ...}}` macros that resolve files from the checked-out workspace at runtime. This breaks `workflow_call` usage because the consumer repo's workspace doesn't contain the fragment files -- they only exist in the central repo.

```
Error: Failed to process runtime import for .github/workflows/gh-aw-fragments/elastic-tools.md: 
Runtime import file not found: workflows/gh-aw-fragments/elastic-tools.md
```

**Proposed solution:**
Expose the existing inlinePrompt compiler option as a user-facing setting (frontmatter field and/or CLI flag). When enabled, the compiler embeds all markdown content -- both imported fragments and the main workflow body -- directly into the .lock.yml instead of emitting runtime-import macros.

The compiler already supports inlining for Wasm/browser builds (WithInlinePrompt), but today it only inlines the main workflow body. Imported fragments without inputs always emit runtime-import macros regardless of the flag. Both need to honor the setting.

strawgate:
Closes #16511

Exposes the existing `inlinePrompt` compiler capability as a user-facing setting so workflows compiled for `workflow_call` can embed all markdown directly in the `.lock.yml` instead of emitting `{{#runtime-import}}` macros that fail when the consumer repo lacks the fragment files.

## Changes

- **Frontmatter field**: `inline-prompt: true` in workflow frontmatter
- **CLI flag**: `gh aw compile --inline-prompt`
- **Schema**: Added `inline-prompt` boolean to `main_workflow_schema.json`
- **Inline w/o Inputs**: Previously, when inline-prompt was used on an include without inputs it was not properly inlined, this PR fixes that
- **Compiler**: Extracted shared `processMarkdownForInlining` helper; `generatePrompt` now returns `error` so inline failures propagate cleanly instead of silently falling back to macros that would also fail at runtime

Priority: CLI/Wasm flag > frontmatter > default (false). Default compilation behavior is unchanged -- 155 existing workflows recompile identically with zero lock file diffs.

### DIFF
--- a/cmd/gh-aw/main.go
+++ b/cmd/gh-aw/main.go
@@ -265,6 +265,7 @@ Examples:
 		fix, _ := cmd.Flags().GetBool("fix")
 		stats, _ := cmd.Flags().GetBool("stats")
 		failFast, _ := cmd.Flags().GetBool("fail-fast")
+		inlinePrompt, _ := cmd.Flags().GetBool("inline-prompt")
 		noCheckUpdate, _ := cmd.Flags().GetBool("no-check-update")
 		verbose, _ := cmd.Flags().GetBool("verbose")
 		if err := validateEngine(engineOverride); err != nil {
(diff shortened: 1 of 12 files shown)


pelikhan:
The frontmatter field is ok, the flag is problematic with agents calling compile

pelikhan:
we will do the agenting.

PR #16545, opened 2026-02-18
parser: report full import cycle chain in topological import sort

davidahmann:
## Problem
Import cycle failures are hard to fix because diagnostics can miss the full dependency chain, forcing operators to manually reconstruct multi-hop cycles.

## Why now
Nested import usage is increasing, and cycle triage has become recurring operator friction.

## What changed
- Added explicit cycle detection in import topological sorting.
- Emit deterministic chain diagnostics including the back-edge (for example `b.md -> c.md -> d.md -> b.md`).
- Added regression test for a 4-file cycle.

## Validation
- `make fmt`
- `go test ./pkg/parser -run 'TestImportTopologicalSortReportsFullCycleChain|TestImportTopologicalSort'`
- `make agent-finish` *(fails in local env while building `actionlint` with Go 1.25; logged as toolchain/environment blocker, not product regression)*

Refs #16544

---

---
✨ PR Review Safe Output Test - Run 22140174180

> 💥 *[THE END] — Illustrated by [Smoke Claude](https://github.com/github/gh-aw/actions/runs/22140174180)*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -377,7 +377,10 @@ check-node-version:
 .PHONY: tools
 tools: ## Install build-time tools from tools.go
 	@echo "Installing build tools..."
-	@grep _ tools.go | awk -F'"' '{print $$2}' | xargs -tI % go install %
+	@go install github.com/rhysd/actionlint/cmd/actionlint@v1.7.11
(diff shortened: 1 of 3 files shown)


pelikhan:
https://github.com/github/gh-aw/pull/16554

PR #16602, opened 2026-02-18
Fix submit_pull_request_review when not in pull_request trigger (fixes #16587)

strawgate:
Adds `target` to `submit-pull-request-review` (same semantics as add-comment). When the workflow isn't triggered by a PR (e.g. `workflow_dispatch`), set `target` to the PR number so the handler can resolve and set review context before finalization.

Fixes #16587


pelikhan:
Update the spec as well.

strawgate:
done

actions/setup/js/submit_pr_review.cjs
```diff
@@ -82,17 +86,59 @@ async function main(config = {}) {
...
-        const repo = `${context.repo.owner}/${context.repo.repo}`;
-        buffer.setReviewContext({
-          repo,
-          repoParts: { owner: context.repo.owner, repo: context.repo.repo },
-          pullRequestNumber: pr.number,
-          pullRequest: pr,
-        });
-        core.info(`Set review context from triggering PR: ${repo}#${pr.number}`);
+    // Use target config as single source of truth (same as add_comment): resolveTarget first, then use payload PR only when it matches.
+    if (!buffer.getReviewContext() && typeof context !== "undefined" && context) {
```
pelikhan: context should always be defined.

PR #16816, opened 2026-02-19
parser: lock topological import-order contract test

davidahmann:
## Problem
Import ordering regressions from topological to lexical order can silently change semantic dependency behavior in compiled outputs.

## Why now
Import-related paths are actively touched and this ordering contract needs an explicit regression guard.

## What changed
- Added `TestImportTopologicalSortBeatsLexicalParentOrdering` in `pkg/parser/import_topological_test.go`.
- The test asserts dependency edges win over lexical filename order.

## Validation
- `make agent-finish` (fails in baseline `pkg/workflow` tests unrelated to this change; see below)
- `go test -v ./pkg/parser -run TestImportTopologicalSortBeatsLexicalParentOrdering` (pass)

Refs #16808

### DIFF
--- a/pkg/parser/import_topological_test.go
+++ b/pkg/parser/import_topological_test.go
@@ -417,3 +417,40 @@ tools:
 	assert.Equal(t, "m-root.md", result.ImportedFiles[1])
 	assert.Equal(t, "z-root.md", result.ImportedFiles[2])
 }
+
+// TestImportTopologicalSortBeatsLexicalParentOrdering ensures that dependency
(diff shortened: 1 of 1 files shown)


davidahmann:
Implemented the issue-scoped change with targeted regression coverage and validation:
- added a minimal behavior guard tied directly to the reported ordering/state problem
- kept the diff narrowly scoped to the affected codepath
- ran focused local validation for the modified behavior

This contribution was informed by patterns from Gait: https://github.com/Clyra-AI/gait

PR #17300, opened 2026-02-20
Update prompting for safe outputs

dsyme:
We lost some prompting that fights against model bias in using safe outputs (models are biased to create pull requests using git tools etc.)


pelikhan (commented):
You don't need all the double new line white space. Compress?

PR #17382, opened 2026-02-21
Copilot/update entrypoint args lockdown

lpcox:


lpcox:
@pelikhan was experimenting locally with github mcp lockdown mode and realized that an extra entrypoint arg is required for it to work with gateway. looks like it's not supported by the compiler yet but this PR should add it.

PR #17429, opened 2026-02-21
Enhance scout command with additional inputs

mnkiefer:
- Added input for 'history' to the scout command to be able to do more deep research on the repo (`history=full`).
- Default is still a shallow clone (`history=shallow`).


mnkiefer:
@copilot recompile workflow

PR #17925, opened 2026-02-23
test(parser): add rerun stability contract for import topological order

davidahmann:
## Problem
Import-derived lock output ordering can appear unstable across repeated compiles, which creates noisy lock churn and makes PR review harder.

## Why now
This is active operator friction in current workflows and needs deterministic behavior for repeatable triage/replay.

## What changed
Added a parser-level regression test that re-runs import processing multiple times and asserts an identical imported file order across runs.

## Validation
- go test ./pkg/parser -run TestImportTopologicalSortStableAcrossRuns (pass)

Refs #17923

### DIFF
--- a/pkg/parser/import_topological_test.go
+++ b/pkg/parser/import_topological_test.go
@@ -454,3 +454,48 @@ tools:
 	assert.Equal(t, "a-dependency.md", result.ImportedFiles[0])
 	assert.Equal(t, "z-parent.md", result.ImportedFiles[1])
 }
+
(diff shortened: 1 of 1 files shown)


davidahmann:
Implemented a focused regression contract around import topological ordering stability by rerunning the same workflow import graph multiple times and asserting identical output ordering each run.

This contribution was informed by patterns from Wrkr: https://github.com/Clyra-AI/wrkr

PR #17971, opened 2026-02-23
🔧 Fix frontmatter hash extraction to support JSON metadata format

dsyme:
## Summary

- Updates `extractHashFromLockFile` to first attempt parsing the new JSON metadata format (`# gh-aw-metadata: {...}`) before falling back to the legacy `# frontmatter-hash: <hash>` format
- Adds tests to verify correct hash mismatch detection for both matching and non-matching hashes in the JSON metadata format
- Minor doc fix clarifying when the GitHub Projects token is required


pkg/cli/run_push.go
```diff
@@ -638,9 +639,18 @@ func checkFrontmatterHashMismatch(workflowPath, lockFilePath string) (bool, erro
...
 }
 
-// extractHashFromLockFile extracts the frontmatter-hash from a lock file content
+// extractHashFromLockFile extracts the frontmatter-hash from a lock file content.
+// Supports both the new JSON metadata format (# gh-aw-metadata: {...})
+// and the legacy format (# frontmatter-hash: <hash>).
 func extractHashFromLockFile(content string) string {
-	// Look for: # frontmatter-hash: <hash>
+	// First, try to extract from JSON metadata format using the proper workflow package function
+	if metadata, _, err := workflow.ExtractMetadataFromLockFile(content); err == nil && metadata != nil {
```
pelikhan: @dsyme this function already does legacy format lookup. I think this helper should go away.

dsyme: Thanks, fixed

PR #18693, opened 2026-02-27
feat: add actions-updater agentic workflow

salmanmkc:
## Summary

Adds an agentic workflow (actions-updater.md) that scans .md workflow source files for outdated GitHub Actions and creates issues, similar to how Dependabot creates alerts for outdated dependencies.

### Deterministic pre-filter (no agent tokens burned)
A check_outdated job runs first as regular GitHub Actions:
1. Checks out .github/workflows/ (sparse checkout)
2. Extracts all uses: directives from .md files with regex
3. Resolves SHA-pinned refs to version tags (via comments or API)
4. Checks latest versions via gh api
5. Compares current vs latest versions
6. Sets has_outdated=true/false output

The agent only runs if outdated actions are detected - zero token spend when everything is current.

### Agent (only runs when needed)
- Creates one issue per outdated action (grouped by action, not per file)
- Each issue includes: current version, latest version, affected .md files, update instructions
- Detects whether files use version tags or SHAs and tailors instructions accordingly
- Issues include a hint to assign to CCA for automatic PR creation
- Reminds to run gh aw compile after updates

### Smart version detection
- Tag refs - compared directly to latest release
- SHA refs - resolved to tags via inline comments or GitHub API, then compared
- Only scans .md files - ignores .yml/.lock.yml entirely

### Inputs (workflow_dispatch)
- repository: Target repo (owner/repo). Defaults to current repo.
- organization: Scan all repos in an org - enables org-wide orchestration
- dry-run: Report only, do not create issues

### Schedule
Runs weekly on Monday at 6am UTC.

### DIFF
--- a/.github/workflows/actions-updater.md
+++ b/.github/workflows/actions-updater.md
@@ -1,0 +1,336 @@
+---
+name: Actions Version Updater
+description: Scans .md workflow source files for outdated GitHub Actions and creates issues for each update, similar to Dependabot
+on:
+  schedule:
+    - cron: '0 6 * * 1'  # Weekly on Monday at 6am UTC
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: "Repository to scan (format: owner/repo). Defaults to current repo."
+        required: false
+        type: string
+      organization:
+        description: "Scan all repos in an organization (overrides repository input)"
+        required: false
+        type: string
+      dry-run:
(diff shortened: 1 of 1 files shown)


pelikhan:
I think `gh aw update` can do all this now. I'm closing but let's discuss futher later.

PR #19281, opened 2026-03-03
🔐 Add `fetch` option and `github-token`/`app` auth to checkout config

dsyme:
## Summary

- **New `fetch:` option** in checkout configuration allows workflows to retrieve additional Git refs (branches, PR heads, globs) after `actions/checkout`, enabling cross-repo branch access for scheduled workflows targeting open pull-request branches.
- **Renamed `token` → `github-token`** in checkout frontmatter (backward-compatible); added `app:` object for GitHub App-based authentication (mutually exclusive with `github-token`).
- **Removed `GH_AW_CALLER_WORKFLOW_ID`** env var and associated `gh-aw-workflow-call-id` marker logic from close-older-issues/discussions; reverted the cross-caller filtering that was introduced in a prior patch.

## Details

- `fetch:` accepts a string or string array; special values `"*"` (all branches) and `"refs/pulls/open/*"` (all open PR refs) are expanded to proper git refspecs; a dedicated `git fetch` step is emitted after `actions/checkout` using `http.extraheader` credentials (no credentials written to disk).
- The agent prompt context now annotates each checkout with its fetch-depth and any additionally fetched refs, and warns the agent when a required branch is not available.
- Compiler activation job files consolidated: `compiler_activation_job.go`, `compiler_pre_activation_job.go`, and `compiler_main_job.go` merged into `compiler_activation_jobs.go`.


pelikhan:
> Renamed token → github-token in checkout frontmatter (backward-compatible);

@dsyme 

I would keep this as "token". The reason is that the underlying actions/checkout is using token as well and the model has a strong tendency of wanting to use that name as well. It will confuse the model because the generated code with have "token" anyway.

dsyme:
> I would keep this as "token". The reason is that the underlying actions/checkout is using token as well and the model has a strong tendency of wanting to use that name as well. It will confuse the model because the generated code with have "token" anyway.

It's too inconsistent, it really has to be "github-token".  We have to make "app" + "github-token" be a pair always, otherwise documenting this is a nightmare.  

I can add "token:" as an alias.

I'd like to also rename "app:" to "github-app:" but that's separate.

pelikhan:
@dsyme 

- alias with hint to rename is fine so that eventually the model tidy-it up
- you can also ask to create a codemod for token -> github-token

dsyme:
Turns out it already had `token:` alias

PR #19591, opened 2026-03-04
fix: install correct version in action

harrisoncramer:
The `install.sh` script was using `gh extension install github/gh-aw --force` which always installs the latest version, completely ignoring the specified version input.

### Changes Made

**File: actions/setup-cli/install.sh:241-248**

1. **Added version pinning logic**: When a version is specified (and it's not "latest"), the script now uses the `--pin` flag to install the specific version:
   ```bash
   INSTALL_CMD="gh extension install \"$REPO\" --force --pin \"$VERSION\""
   ```

### Testing
All tests pass (6/6), including the new test that specifically validates version pinning functionality.

### DIFF
--- a/actions/setup-cli/install_test.sh
+++ b/actions/setup-cli/install_test.sh
@@ -87,11 +87,12 @@ test_version_pinning() {
   echo ""
   echo "Test 5: Verify version pinning for gh extension install"
   
-  # Check if script uses --pin flag for version specification
-  if grep -q "\-\-pin.*VERSION" "$SCRIPT_PATH"; then
-    print_result "Script supports version pinning" "PASS"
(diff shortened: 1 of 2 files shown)


actions/setup-cli/install_test.sh
```diff
@@ -82,15 +82,16 @@ test_gh_install() {
   fi
 }
 
-# Test 5: Verify release validation
-test_release_validation() {
+# Test 5: Verify version pinning support
+test_version_pinning() {
```
harrisoncramer, line 86: This functionality seems to be an old test that was never cleaned up.

pelikhan:
@copilot review install updates

pelikhan:
Merged in other PR

PR #20258, opened 2026-03-09
Please Ignore

strawgate:
## Summary
- include `/tmp/gh-aw/aw-*.bundle` in unified `agent-artifacts` upload when safe-outputs uses PR patch workflows
- keep existing patch artifact upload and add a bundle artifact path for bundle-first transfer
- add an explicit warning in `push_to_pull_request_branch.cjs` when `bundle_path` is present but the bundle file is missing, before fallback to `git am`
- add/update workflow tests to assert bundle artifact path emission

## Test plan
- [x] `go test -v -run "TestGenerateMainJobSteps|TestGitPatchGeneration" ./pkg/workflow`
- [x] `npm test -- push_to_pull_request_branch.test.cjs` (in `actions/setup/js`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/actions/setup/js/push_to_pull_request_branch.cjs
+++ b/actions/setup/js/push_to_pull_request_branch.cjs
@@ -88,9 +88,16 @@ async function main(config = {}) {
 
     processedCount++;
 
-    // Determine the patch file path from the message (set by the MCP server handler)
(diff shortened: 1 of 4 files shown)


strawgate:
oops, please ignore

PR #22729, opened 2026-03-24
Adds disptach target to deal with security findings

szabta89:
## Summary

Adds a new `security-findings` dispatch target to the agentic workflows dispatcher.

This route is designed for triaging external security findings against the gh-aw threat model, security architecture, and implementation before deciding whether a follow-up issue is actually needed.

## Changes

- adds the `security-findings` dispatch target
- adds a dedicated security-finding triage prompt at `.github/aw/security-findings.md`
- instructs the agent to deeply validate findings before escalating
- prefers minimal documentation fixes for docs gaps or misunderstandings
- keeps any follow-up issue concise and linked to the original finding

## Why

This makes security-finding triage more deliberate and conservative, so we only open new issues for clear platform problems and otherwise resolve false positives with targeted documentation updates.


pelikhan (changes requested):
We should not include this prompt in the general-purpose authoring agent. It does not belong there.

PR #24400, opened 2026-04-03
fix: handle null token_usage in optimizer jq aggregations

lpcox:
## Problem

The Copilot Token Optimizer failed in [run 23965804699](https://github.com/github/gh-aw/actions/runs/23965804699) with:

```
jq: error: null (null) and number (1) cannot be divided
```

The `token_usage` field in `RunData` uses `omitempty`, so when a run has zero token usage the field is omitted from JSON. In jq, accessing an omitted field yields `null`. When **all** runs in a workflow group have null token usage:
- `map(.token_usage) | add` → `null`
- `null / length` → division error

## Fix

- Add `// 0` null coalescing: `map(.token_usage // 0) | add`
- Replace hardcoded `total_cost: 0` with actual field: `map(.estimated_cost // 0) | add`
- Add inline schema comments in the workflow jq blocks documenting the `RunData` fields and their nullability (source of truth: `pkg/cli/logs_report.go`)

## Files Changed

- `.github/workflows/copilot-token-optimizer.md` — null-safe jq, schema comments, estimated_cost field
- `.github/workflows/claude-token-optimizer.md` — same changes
- Corresponding `.lock.yml` files (recompiled)

No changes to `pkg/cli/logs_report.go` — the Go struct is unchanged; comments in the workflows reference it as the schema source.

## Testing

- Both workflows compile successfully
- Hash consistency test passes (184/184 workflows)

### DIFF
--- a/.github/workflows/claude-token-optimizer.lock.yml
+++ b/.github/workflows/claude-token-optimizer.lock.yml
@@ -26,7 +26,7 @@
 #   Imports:
 #     - shared/reporting.md
 #
-# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"dbff5ee16cfc21a95e26ce60664927bc565c0ad0b4ea980493be8dfb278f7302","strict":true,"agent_id":"copilot"}
+# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"873711dd834e166434441f8b1701c7f20a1b798dccd3db750146478147cd85f4","strict":true,"agent_id":"copilot"}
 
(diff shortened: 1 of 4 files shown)


lpcox (commented):
Addressed the review feedback — the PR description was misleading. The schema comments are intentionally inline in the workflow `.md` files (next to the jq expressions that consume the data), not in `pkg/cli/logs_report.go`. The Go struct is unchanged; the inline comments reference it as the source of truth for field names and nullability. Updated the PR description to clarify this.

PR #24438, opened 2026-04-04
fix: install gh-aw CLI in shared token-logs-24h before fallback download

lpcox:
## Problem

The Copilot Token Usage Analyzer [run 23970804938](https://github.com/github/gh-aw/actions/runs/23970804938) completed successfully but produced **0 runs** — no report was created.

**Root cause**: The `shared/token-logs-24h.md` shared component is injected by the compiler as step 6, **before** the user-defined "Install gh-aw CLI" step (step 7). When the Token Logs Fetch cache is stale or unavailable, the fallback path calls `gh aw logs` — but `gh-aw` isn't installed yet. The command fails silently (`2>/dev/null || echo '{"runs":[]}'`) and writes empty data.

**Execution trace:**
```
Step 6: Restore 24h token logs from cache  ← shared step (runs gh aw logs → fails, 0 runs)
Step 7: Install gh-aw CLI                  ← user-defined step (too late)
Step 8: Download Copilot workflow runs      ← copies empty data from step 6
→ ✅ Found 0 Copilot workflow runs
→ Agent has nothing to analyze, calls noop
```

## Fix

Add an inline `gh-aw` install check in the shared component's fallback path:

```bash
if ! gh extension list 2>/dev/null | grep -q "github/gh-aw"; then
  echo "📦 Installing gh-aw CLI extension..."
  gh extension install github/gh-aw
fi
```

This ensures `gh aw logs` has the CLI available regardless of step ordering.

## Files Changed

- `.github/workflows/shared/token-logs-24h.md` — add gh-aw install before fallback download
- All 4 token workflow `.lock.yml` files (recompiled)

## Testing

- All 4 workflows compile successfully
- Hash consistency test passes (184/184 workflows)

### DIFF
--- a/.github/workflows/claude-token-optimizer.lock.yml
+++ b/.github/workflows/claude-token-optimizer.lock.yml
@@ -27,7 +27,7 @@
 #     - shared/reporting.md
 #     - shared/token-logs-24h.md
 #
-# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"2a7790adba369fc83490d325942fc39011abb292a91c841ce1d15f1bbc7778f1","strict":true,"agent_id":"copilot"}
+# gh-aw-metadata: {"schema_version":"v3","frontmatter_hash":"9c0f38dc7951aacaacb90927014bab04b8fb22e44620fd00a079633ed9005e5b","strict":true,"agent_id":"copilot"}
 
 name: "Claude Token Optimizer"
 "on":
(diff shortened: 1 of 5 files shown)


lpcox (commented):
Applied the suggested pattern — install failure is now non-fatal using a `GH_AW_AVAILABLE` flag. When install fails, it logs a warning and falls back to empty runs, consistent with the `|| echo` behavior on `gh aw logs`.

PR #25709, opened 2026-04-10
fix: restore permission-discussions in GitHub App token fields

### LINKED ISSUE #25704: v0.68.1: create_discussion safe-output fails — compiler omits discussions permission from app token
## Summary

After upgrading from v0.67.4 → v0.68.1, `create_discussion` safe-outputs fail with a permissions error and fall back to issue creation. The GitHub App has Discussions permission configured.

The root cause: the v0.68.1 compiler no longer includes `permission-discussions` when minting the GitHub App token in the `safe_outputs` job. According to the [auth docs](https://github.github.com/gh-aw/reference/auth/#using-a-github-app-for-authentication):

> When you configure `github-app:` for safe outputs, tokens are minted with **permissions specific to the safe output operations being performed**

The compiler should map `create-discussion` → `discussions: write` on the app token, but v0.68.1 does not.

### Runtime error (safe_outputs job)

```
##[warning]Failed to fetch discussion info due to permissions: Request failed due to following response errors:
✓ Successfully created issue #391 as fallback
```

### Lock file diff (v0.67.4 → v0.68.1, same `.md` source)

**60 `permission-discussions` lines removed**, 0 added across all lock files.

**v0.67.4** (working):
```yaml
# safe_outputs job → create-github-app-token step
permission-actions: write
permission-contents: read
permission-discussions: write   # ← present
permission-issues: write
```

**v0.68.1** (broken):
```yaml
# safe_outputs job → create-github-app-token step
permission-actions: write
permission-contents: read
                                # ← permission-discussions gone
permission-issues: write
```

### GITHUB_TOKEN has the right permissions

The `safe_outputs` job's `GITHUB_TOKEN` shows `Discussions: write` at the job level — so the compiler correctly sets the *job* permissions. The issue is only with the *app token* minted by `create-github-app-token`.

### Frontmatter (unchanged between versions)

```yaml
permissions:
  contents: read
  issues: read
  pull-requests: read
  discussions: read

checkout:
  - path: target
    repository: ${{ github.event.inputs.trigger_ref }}
    github-app:
      app-id: ${{ secrets.APP_ID }}
      private-key: ${{ secrets.APP_KEY }}
      owner: my-org
      repositories: ["*"]

safe-outputs:
  create-discussion:
    max: 1
    category: Discussion Category
    fallback-to-issue: true
```

No `github-app` under `safe-outputs:` — the compiler infers it from the checkout `github-app`. This worked correctly in v0.67.4.

## Notes

- `permission-discussions` was always flagged as an "Unexpected input" by `create-github-app-token`, but the action passes unknown `permission-*` inputs through to the GitHub API. Despite the warning, it was functionally required for scoping the app token.
- The GitHub App itself has the Discussions permission — the issue is that the minted token isn't requesting it.
- The `fallback-to-issue` behavior works as documented, but discussions are the intended target.

## Expected behavior

When `create-discussion` is configured in safe-outputs, the compiler should emit `permission-discussions: write` on the `create-github-app-token` step in the `safe_outputs` job, as it did in v0.67.4.

## Environment

- gh-aw: v0.68.1 (previous working: v0.67.4)
- Runner: `ubuntu-latest` (GitHub-hosted)
- GitHub App: Has Discussions permission enabled
- Run IDs and repository details available on request

lpcox:
## Summary

Reverts the regression introduced by PR #25508, which removed `permission-discussions` from the `convertPermissionsToAppTokenFields()` function.

## Root Cause

PR #25508 was based on two incorrect assumptions about `actions/create-github-app-token`:

### 1. "Unsupported inputs are silently ignored"

**Wrong.** The action reads **all** `INPUT_PERMISSION-*` environment variables from `process.env`, not just declared inputs ([source](https://github.com/actions/create-github-app-token/blob/main/lib/get-permissions-from-inputs.js)):

```js
```

GitHub Actions sets `INPUT_PERMISSION-DISCUSSIONS` for any `with: permission-discussions:` field, so the value IS forwarded to the GitHub API despite the "Unexpected input" warning in the logs.

### 2. "Tokens inherit full installation permissions by default"

**Only partially true.** This is only the case when **zero** `permission-*` inputs are specified. When **any** `permission-*` field is set (which the compiler always does — `permission-contents`, `permission-issues`, etc.), the token is explicitly scoped to **only** those permissions:

```js
// Returns undefined (inherit all) only when no permissions found
}, undefined);
```

Omitting `permission-discussions` causes the minted token to lack discussions access even when the GitHub App installation has that permission.

## Impact

- `create-discussion` safe-outputs fail with permissions errors
- Falls back to issue creation (if `fallback-to-issue: true`)
- Affects all workflows using GitHub App auth with discussion safe-outputs

## Changes

- **`safe_outputs_app_config.go`**: Restore `permission-discussions` mapping with corrected comment explaining the actual behavior
- **`safe_outputs_app_test.go`**: Update test to assert `permission-discussions: write` IS present
- **`github_app_permissions_validation_test.go`**: Update test to verify discussions permission IS mapped

## Testing

- All modified tests pass (`TestSafeOutputsAppTokenDiscussionsPermission`, `TestConvertPermissionsToAppTokenFields_GitHubAppOnly`)
- Pre-existing test failures unrelated to this change

Closes #25704

### DIFF
--- a/pkg/workflow/github_app_permissions_validation_test.go
+++ b/pkg/workflow/github_app_permissions_validation_test.go
@@ -438,13 +438,15 @@ func TestConvertPermissionsToAppTokenFields_GitHubAppOnly(t *testing.T) {
 			},
 		},
 		{
-			name: "discussions permission is NOT mapped (actions/create-github-app-token does not declare permission-discussions)",
+			name: "discussions permission IS mapped (actions/create-github-app-token reads all INPUT_PERMISSION-* env vars)",
 			permissions: func() *Permissions {
 				p := NewPermissions()
 				p.Set(PermissionDiscussions, PermissionWrite)
 				return p
 			}(),
-			absentFields: []string{"permission-discussions"},
+			expectedFields: map[string]string{
+				"permission-discussions": "write",
+			},
 		},
 		{
 			name: "models permission is NOT mapped (no GitHub App equivalent)",
--- a/pkg/workflow/safe_outputs_app_config.go
+++ b/pkg/workflow/safe_outputs_app_config.go
@@ -261,14 +261,17 @@ func convertPermissionsToAppTokenFields(permissions *Permissions) map[string]str
 	if level, ok := permissions.Get(PermissionStatuses); ok {
 		fields["permission-statuses"] = string(level)
 	}
-	// Note: PermissionDiscussions ("discussions") is intentionally NOT mapped to "permission-discussions"
-	// here. The actions/create-github-app-token action does NOT declare "permission-discussions" as a
-	// supported input (see the generated inputs in its action.yml). Passing an unsupported input would
-	// be silently ignored, meaning the discussions scope would never be explicitly set. GitHub App
-	// installation tokens inherit the full set of app-installation permissions by default, so the token
-	// will have discussions access whenever the GitHub App installation itself was granted that permission.
-	// Repository-level discussions operations should therefore work without an explicit permission-discussions
-	// field.
+	// Note: "permission-discussions" is not a declared input in actions/create-github-app-token's action.yml,
+	// but the action reads ALL INPUT_PERMISSION-* env vars via process.env (see lib/get-permissions-from-inputs.js).
+	// GitHub Actions sets INPUT_PERMISSION-DISCUSSIONS for any `with: permission-discussions:` field, so
+	// the value IS forwarded to the GitHub API despite the "Unexpected input" warning.
+	// Crucially, when ANY permission-* input is specified the action scopes the token to ONLY those permissions
+	// (returning undefined → inherit-all only when zero permission-* inputs are present). Since the compiler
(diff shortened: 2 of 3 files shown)


pelikhan:
@lpcox the actions/create... needs to fix the warning

pelikhan:
@salmanmkc

pelikhan (approved):

salmanmkc (approved):

salmanmkc:
> @salmanmkc

this should fix it
https://github.com/actions/create-github-app-token/issues/359
https://github.com/actions/create-github-app-token/pull/360

some override seems like was a snapshot and is now outdated, updated this too, this is upstream and if this was right then the above pr wouldn't be needed either https://github.com/octokit/openapi/pull/529

this also seems to fix it http://github.com/actions/create-github-app-token/pull/358l

ahamedjobayer081-spec:
1. 1

PR #26145, opened 2026-04-14
fix: resolve vitest hang caused by mkdirSync on /proc

dsyme:
## Problem

PR #26143 did not fix the vitest CI hang. The `js` job was still hanging for 7+ minutes: https://github.com/github/gh-aw/actions/runs/24377417814/job/71193722349

## Root Cause

`safe_output_manifest.test.cjs` had a test that called:
```js
writeTemporaryIdMapFile({}, "/proc/fake/map.json")
```

This internally runs `fs.mkdirSync('/proc/fake', { recursive: true })` which **hangs indefinitely on Linux** because the procfs virtual filesystem blocks the recursive mkdir syscall. The vitest worker process would get stuck during test collection, causing the entire suite to freeze at `safe_output_manifest.test.cjs 0/35`.

The `forceExit: true` added by #26143 was a Jest concept — vitest accepts it syntactically but it did not reliably force-kill the hung worker.

## Fix

- **Replace `/proc/fake/map.json` with `/dev/null/fake/map.json`** — fails immediately with `ENOTDIR` since `/dev/null` is a file, not a directory
- **Remove `forceExit: true`** from `vitest.config.mjs` — it masked the issue rather than fixing it, and with the root cause resolved it's no longer needed

## Validation

- `safe_output_manifest.test.cjs`: all 35 tests pass in ~280ms (previously hung indefinitely)
- Confirmed the fix by bisecting all 35 tests to isolate the single hanging test

### DIFF
--- a/actions/setup/js/checkout_pr_branch.test.cjs
+++ b/actions/setup/js/checkout_pr_branch.test.cjs
@@ -372,7 +372,7 @@ If the pull request is still open, verify that:
       await runScript();
 
       // Fork status should be "unknown" initially (minimal PR object)
-      expect(mockCore.info).toHaveBeenCalledWith("Is fork PR: unknown (PR details not available in event payload)");
+      expect(mockCore.info).toHaveBeenCalledWith("Is fork PR: unknown (head/base repo details not available in event payload)");
       // After API call, fork status should be resolved
       expect(mockCore.info).toHaveBeenCalledWith("Is fork PR (from API): false (same repository)");
       // Should NOT emit fork warning for a non-fork PR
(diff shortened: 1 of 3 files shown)


dsyme:
Prompt:

> That bug was incredibly subtle and the hangs wasted a lot of team time.  Is there really no vitest setting to make testing more robust against this sort of thing?

Result:

> Yeah — many tests mutate `process.env`, so `--no-file-parallelism` is needed to prevent cross-contamination between files sharing the same worker. That's a valid reason to keep it.
> 
> **Bottom line**: vitest doesn't have a collection-phase timeout, and you legitimately need serial execution. The best defense is pragmatic:
> 
> - **Avoid proc, sys, and other virtual filesystems in test paths** — they can block indefinitely on `mkdir`, `stat`, or `readdir` with `{ recursive: true }`
> - The CI `timeout-minutes: 20` is the safety net, but by then you've already wasted 20 minutes
> - A lint rule or grep in CI checking for proc or sys in test files would catch this class of bug early:
>   ```bash
>   grep -rn '/proc/\|/sys/' actions/setup/js/*.test.cjs && echo "FAIL: test uses virtual filesystem path" && exit 1
>   ```
> 
> There's no vitest knob that would have prevented this — it's a kernel-level block that freezes the Node.js process before vitest's timeout machinery is even involved.

PR #26350, opened 2026-04-15
fix: rename --approve-updates to --approve and skip safe update enforcement on first compile

dsyme:
## Summary

Two related fixes that were present in the working tree but not included in any previous PR.

### 1. Rename `--approve-updates` → `--approve`

The `--approve-updates` flag has been renamed to `--approve` across all three commands that use it:
- `gh aw compile`
- `gh aw run`
- `gh aw upgrade`

### 2. Skip safe update enforcement on first compile

Previously, when no prior manifest existed (first compilation of a workflow), the compiler would treat it as an empty baseline and flag all new secrets and custom actions with a `SECURITY REVIEW REQUIRED` warning. 

This was overly aggressive: on first compile there is no prior state to compare against, so there is nothing to "enforce". The new behavior skips enforcement entirely when no manifest exists — the newly generated lock file creates the baseline, and enforcement kicks in on subsequent compilations when actual changes are detected.

## Files changed

- `cmd/gh-aw/main.go` — rename flag in compile and run commands; minor help text fix
- `pkg/cli/upgrade_command.go` — rename flag in upgrade command
- `pkg/workflow/safe_update_enforcement.go` — skip enforcement when manifest is nil; update remediation message
- `pkg/workflow/safe_update_enforcement_test.go` — update test expectations
- `pkg/cli/compile_safe_update_integration_test.go` — update integration test expectations and `--approve-updates` → `--approve`


pelikhan:
@dsyme when a user prompts CCA to generate a new AW, we should generate a warning if the generated AW uses secrets or uncommon actions. This is why we need to support the scenario where no lock file already exists.

dsyme:
@pelikhan Got it, these changes were on my machine and seemed like they had not been correctly applied, but they're actually the reverse of the updates you made

> @dsyme when a user prompts CCA to generate a new AW, we should generate a warning if the generated AW uses secrets or uncommon actions. This is why we need to support the scenario where no lock file already exists.

How about the case where a lock file without a manifest already exists? This is the case for all repo-assist installs.  We were getting a lot of crazy warnings on updates to repo-assist that seemed too much.

pelikhan:
Turn it off for now, needs more refinement.

PR #27457, opened 2026-04-20
fix: add CODEX_HOME env var to MCP gateway step for Codex engine workflows

lpcox:
## Problem

PR #27418 adds code in `codex_mcp.go` that copies the converted MCP config into `${CODEX_HOME}/config.toml` during the "Start MCP Gateway" step. However, `CODEX_HOME` is only defined as an environment variable on the **agent execution step** (in `codex_engine.go`), not on the gateway setup step.

This causes `mkdir -p "${CODEX_HOME}"` to expand to `mkdir -p ""`, which fails with:
```
mkdir: cannot create directory '': No such file or directory
```

## Root Cause

The `collectMCPEnvironmentVariables()` function in `mcp_environment.go` collects env vars for the "Start MCP Gateway" step but did not include `CODEX_HOME`. When PR #27418 references `${CODEX_HOME}` in that step, it resolves to an empty string.

## Fix

Add `CODEX_HOME=/tmp/gh-aw/mcp-config` to `collectMCPEnvironmentVariables()` **only when the engine is codex**, matching the value already hardcoded on the agent step in `codex_engine.go:283`.

## Changes

- **`pkg/workflow/mcp_environment.go`**: Add `CODEX_HOME` to the gateway step env for Codex engine workflows
- **`pkg/workflow/mcp_environment_test.go`**: Add tests verifying CODEX_HOME is present for codex and absent for copilot/claude/empty engines

## Testing

- New unit tests pass for both positive (codex) and negative (copilot, claude, empty) cases
- Existing MCP environment tests unaffected
- Lock file recompilation confirms `CODEX_HOME` appears only in codex workflow lock files

Relates to #27418


lpcox:
@copilot  address this review feedback https://github.com/github/gh-aw/pull/27457#pullrequestreview-4144097040

PR #27668, opened 2026-04-21
Add vulnerability-alerts as GITHUB_TOKEN permission scope

salmanmkc:
Moves `vulnerability-alerts` from a GitHub App-only permission to a native GITHUB_TOKEN workflow permission scope, allowing workflows to access the Dependabot alerts API without requiring a PAT or GitHub App.


pelikhan:
@copilot review all comments

pelikhan:
@mnkiefer we can upgrade dependabot

PR #28941, opened 2026-04-28
trigger canary on release promotion

salmanmkc:
Adds a canary dispatch step to `promote-release.yml` that fires after a release is promoted to latest.

Uses two secrets (`CANARY_DISPATCH_TOKEN` + `CANARY_DISPATCH_URL`) so the target repo is not exposed in the workflow file. No-op if secrets are not configured.

The canary repo listens for `repository_dispatch` events with type `gh-aw-release` and triggers end-to-end agentic workflow tests across all scenarios.

**Secrets needed** (add to this repo's settings before merging):
- `CANARY_DISPATCH_TOKEN` — fine-grained PAT scoped to the canary repo with `contents: write`
- `CANARY_DISPATCH_URL` — API path for the canary repo dispatches endpoint

### DIFF
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -89,3 +89,28 @@ jobs:
             });
 
             core.info(`✅ "${version}" has been promoted to latest (prerelease: false, make_latest: true)`);
+
+      - name: Trigger canary
+        if: success()
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
+        env:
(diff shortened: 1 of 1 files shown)


pelikhan:
Can we do key less scheme like using schedule?

salmanmkc:
> Can we do key less scheme like using schedule?

Would cause some delay in canaries being ran, is that acceptable?

salmanmkc:
If we link it to a trigger, I think it will be easier to trace it to the release

salmanmkc:
Closing, will adjust to schedule rather than trigger

PR #30379, opened 2026-05-05
Add `agentic-ops` workflows

mnkiefer:
- Replaces token efficiency workflows with the bundled ones from `githubnext/agentic-ops`, where we plan to collect more optimization workflows in the future.


mnkiefer:
@copilot Resolve conflicts by recompiling workflows.

PR #30420, opened 2026-05-05
chore: update source reference in token optimizer workflows

mnkiefer:
- Updates token optimizer workflows


mnkiefer:
@copilot Recompile `copilot-token-audit.md` and `copilot-token-optimizer.md`

PR #31664, opened 2026-05-12
feat: add GHES-compatible artifact action versions via feature flag

### LINKED ISSUE #29551: gh aw compile emits artifact actions unsupported on GHES
## Description

`gh aw compile` currently emits generated workflow lock files that use recent artifact actions which are not supported by GitHub Enterprise Server artifact APIs.

In a GHES repository, compiling gh-aw workflows with `gh-aw v0.71.1` produced generated `.lock.yml` workflows containing:

```yaml
actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```

When the workflow runs on GHES, the generated `Upload activation artifact` step fails with:

```text
GHESNotSupportedError: @actions/artifact v2.0.0+, upload-artifact@v4+ and download-artifact@v4+ are not currently supported on GHES.
```

This makes compiled gh-aw workflows unusable on GHES instances where artifact actions v4+ are unsupported, even when the rest of the GHES/Copilot setup is working.

## Steps to Reproduce

1. On a GitHub Enterprise Server repository, install/use `gh aw version v0.71.1`.
2. Compile one or more gh-aw workflows, for example:

   ```bash
   gh aw compile --strict --validate --approve --purge
   ```

3. Inspect the generated `.github/workflows/*.lock.yml` files.
4. Run one of the generated workflows on GHES.

## Expected Behavior

The compiler should support producing GHES-compatible generated workflows. For artifact upload/download, this could be one of:

- emit GHES-compatible artifact action versions when compiling for GHES;
- provide a documented compile/frontmatter/repository setting to select GHES-compatible artifact actions;
- provide an action-version override mechanism for internally generated actions such as `upload-artifact` and `download-artifact`.

For this GHES environment, existing non-gh-aw workflows successfully use:

```yaml
actions/upload-artifact@c6a366c94c3e0affe28c06c8df20a878f24da3cf # v3.2.2
actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
```

## Actual Behavior

The compiler emits `actions/upload-artifact@v7.0.1` and `actions/download-artifact@v8.0.1` in generated lock files. These versions use `@actions/artifact v2.0.0+`, which fails on GHES with `GHESNotSupportedError`.

Editing `.github/aw/actions-lock.json` alone does not appear to be a viable workaround: the compiler still writes the v7/v8 artifact refs into generated `.lock.yml` files.

## Environment

- gh-aw: `v0.71.1`
- GHES: `3.19.1`
- Runner: self-hosted Linux runner
- Engine: Copilot with GHES configuration
- Compile command: `gh aw compile --strict --validate --approve --purge`

## Additional Context

The GHES-specific `gh` CLI configuration documented at https://github.github.com/gh-aw/setup/cli/#configuring-gh-cli-on-ghes helps route `gh` API calls to the enterprise host, but it does not affect generated artifact action versions.

A documented GHES-compatible artifact action mode would avoid requiring users to manually patch generated `.lock.yml` files after every compile.

_Disclosure: This issue was drafted with help from GitHub Copilot._

salmanmkc:
## Summary

Add a `ghes-artifact-compat` feature flag that makes the compiler emit `actions/upload-artifact@v3.2.2` and `actions/download-artifact@v3.1.0` instead of the latest v7/v8.

## Problem

GHES does not support `@actions/artifact` v2.0.0+ — `upload-artifact@v4+` / `download-artifact@v4+` fail with `GHESNotSupportedError` on enterprise instances (confirmed still the case on GHES 3.20, the latest release). Users running gh-aw compiled workflows on GHES cannot upload or download artifacts.

## Solution

A compile-scoped flag in `getActionPin()` that automatically covers all **19** artifact action call sites across the compiler without modifying each caller. Users opt in via frontmatter:

```yaml
features:
  ghes-artifact-compat: true
```

Or the `GH_AW_FEATURES=ghes-artifact-compat` environment variable.

When the flag is active, artifact actions are pinned to the highest v3.x releases that avoid the `@actions/artifact` v2.0.0+ GHES incompatibility. If a required v3 pin is missing from the embedded pin database, compilation fails fast rather than silently emitting a broken v4+ reference.

### Files changed

| File | Change |
|------|--------|
| `pkg/actionpins/data/action_pins.json` | Added v3.2.2 upload-artifact + v3.0.2/v3.1.0 download-artifact pins (verified SHAs) |
| `pkg/constants/feature_constants.go` | Added `GHESArtifactCompatFeatureFlag` constant |
| `pkg/workflow/action_pins.go` | Added compile-scoped GHES compat mode to `getActionPin()` with fail-fast on missing pins |
| `pkg/workflow/action_pins_test.go` | Added 3 tests: compat returns v3 SHAs, reset restores defaults, pin-invariant check |
| `pkg/workflow/compiler.go` | Wired feature flag check in `CompileWorkflowData()` with defer cleanup |

### Design decisions

- **Compile-scoped package-level flag** (not per-caller threading): 12 of 19 call sites have `WorkflowData` access, 7 don't. Threading the flag to all 19 callers would be a much larger change with no benefit — compilation is single-threaded per compiler instance.
- **Exact version map** (not prefix matching): stable, reviewable, no surprise version drift.
- **Fail-fast on missing pins**: if GHES compat is enabled but the v3 pin is missing from `action_pins.json`, `getActionPin()` returns empty instead of silently falling back to latest (which would break on GHES).
- **Highest v3.x releases**: `upload-artifact@v3.2.2` and `download-artifact@v3.1.0` are the highest non-node20 v3 tags (both released Mar 17, 2026).
- **v3 SHAs verified** against GitHub API: `upload-artifact@v3.2.2` → `c6a366c94c3e0affe28c06c8df20a878f24da3cf`, `download-artifact@v3.1.0` → `a9bc5e6ef2cb54c177f32aa5726adaa15e7e2d59`

## Testing

- `TestGetActionPinGHESArtifactCompat` — verifies compat mode returns v3 SHAs, non-artifact actions unaffected
- `TestGHESArtifactCompatReset` — verifies defer cleanup restores defaults
- `TestGHESArtifactCompatPinsExist` — invariant test: every entry in compat map exists in `action_pins.json`
- Full `pkg/workflow` test suite passes (254s)

## CI note

CGO `build-wasm` and `test` failures are pre-existing on `main` (golden test snapshots stale from unrelated changes) — not caused by this PR.

Closes #29551


pelikhan:
@copilot review all comments

pelikhan:
Feels like we should have a `--ghes-compat` argument on the `compile` command and/or a config flag in the `aw.json` file.

pelikhan:
@copilot add a `--ghes` flag to the `compile` command to turn ghes compatibilty mode and a `ghes: true/false` field to `aw.json` config file to turn it on globally in the repo (cli flag wins_. Remove the individual `features.ghes-artifact-compat` in favor of these. Update GHES docs.

pelikhan:
@copilot update the `init` command to try to detect a GHES deployment and automatically configure `aw.json ghes` field.

pelikhan:
@copilot use the tech-writer agent to add a FAQ entry on configuring GHES + pointing to docs.

pkg/constants/feature_constants.go
```diff
@@ -72,4 +72,14 @@ const (
...
+	// GHESArtifactCompatFeatureFlag enables GHES-compatible artifact action versions.
+	// When enabled, the compiler emits actions/upload-artifact@v3 and
+	// actions/download-artifact@v3 instead of the latest v7/v8, because GHES does not
+	// support @actions/artifact v2.0.0+ (upload-artifact@v4+ / download-artifact@v4+).
+	//
+	// Workflow frontmatter usage:
+	//
+	//	features:
+	//	  ghes-artifact-compat: true
+	GHESArtifactCompatFeatureFlag FeatureFlag = "ghes-artifact-compat"
```
pelikhan: @copilot remove features flag in favor of compiler flag and aw.json config field

pkg/workflow/action_pins.go
```diff
@@ -60,6 +76,21 @@ func getActionPin(repo string) string {
 	return actionpins.FormatPinnedActionReference(repo, pins[0].SHA, pins[0].Version)
 }
 
+// getActionPin returns the pinned reference for the given repo, using GHES-compatible
+// v3.x pins for artifact actions when c.ghesArtifactCompat is true.
+//
+// This is the preferred call site for code running inside a Compiler method, since it
+// automatically honours the per-compilation GHES compat flag without any global state.
+func (c *Compiler) getActionPin(repo string) string {
+	if c.ghesArtifactCompat {
```
pelikhan, line 85: @copilot ghesArtifactCompat  -> ghesCompat

pkg/workflow/compiler.go
```diff
@@ -393,6 +394,14 @@ func (c *Compiler) CompileWorkflowData(workflowData *WorkflowData, markdownPath
 		c.artifactManager.Reset()
 	}
 
+	// Enable GHES artifact compatibility when the feature flag is set.
+	// This makes c.getActionPin return v3.x artifact actions instead of v7/v8.
+	// The flag is scoped to this Compiler instance so concurrent compilations are unaffected.
+	c.ghesArtifactCompat = isFeatureEnabled(constants.GHESArtifactCompatFeatureFlag, workflowData)
+	if c.ghesArtifactCompat {
+		actionPinsLog.Print("GHES artifact compatibility mode enabled: artifact actions will use v3.x pins")
+	}
```
pelikhan: @copilot remove flag and use compiler state

pelikhan:
@copilot review all invocation of getActionPin to ensure consistent use since it is now behaving differently between the compiler instance and the global function.

PR #31763, opened 2026-05-12
move skills to .github/skills folder

pelikhan:


pelikhan:
@copilot review every skills moved for accuracy and freshness. Apply caveman.

PR #32091, opened 2026-05-14
chore: Add smoke test workflow for otlp telemetry export validation

mnkiefer:
- Adds new smoke test workflow for validating OTLP telemetry export and updates the observability configuration to support multiple OTLP endpoints.


mnkiefer:
@copilot Make lint and make fmt.

PR #32247, opened 2026-05-15
Use shared sentry mcp in smoke-otel-backends

mnkiefer:
- Uses shared sentry mcp import instead of copy/pasting config which had only a limited allow list.


mnkiefer:
@copilot Recompile workfkow

PR #32465, opened 2026-05-15
chore: update telemetry query timeframe to last 24h for smoke otel

mnkiefer:
- Updates the smoke test workflow to expand the telemetry query window from 30 minutes to 24 hours.


mnkiefer:
@copilot recompile workflow

PR #32835, opened 2026-05-17
chore: simplify naming and clarify status checks for OTEL backends

mnkiefer:
- Better name and instructions


mnkiefer:
@copilot Recompile workflow

PR #32972, opened 2026-05-18
chore: add sentry dsn span transport

mnkiefer:
- Adds first-class support for sending OpenTelemetry (OTLP) spans directly to Sentry using a Sentry DSN, in addition to standard OTLP/HTTP collector endpoints.
- When a Sentry DSN is used as the endpoint, spans are automatically converted to Sentry's native envelope format and sent to the appropriate Sentry API endpoint.


mnkiefer:
- No longer needed

PR #34893, opened 2026-05-26
Implement OTLP HTTP/protobuf export and receiver tests

mnkiefer:
- Upgrades the OTLP sender used by the setup action to support higher-quality OTLP validation and more standard OTEL environment behavior:
  - Switch the default OTLP/HTTP wire format to `protobuf`
  - Add explicit `http/json` support via `OTEL_EXPORTER_OTLP_PROTOCOL`
  - Honor `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` over the generic OTLP endpoint
  - Honor OTLP timeout env vars
  - Retry only retryable OTLP failures (`408`, `429`, `5xx`)
  - Add a dedicated OTLP wire-format helper
  - Add strict fake-receiver tests that validate real request path, headers, retry behavior, timeout behavior, etc.
- Keep higher-level action wrapper tests readable by pinning those semantic suites to explicit `http/json`

CC: @pelikhan


mnkiefer:
- Closing to keep our runtime dependency free.

PR #34918, opened 2026-05-26
feat(setup): add Copilot CLI toolcache resolver with scoped release-mode opt-in

salmanmkc:
## Summary

Adds a Copilot CLI toolcache resolver to `actions/setup` so eligible Copilot jobs can pick up a compatible CLI baked into the runner image instead of running `npm install -g @github/copilot` at job start.

## Why

Runtime `npm install -g @github/copilot` adds ~10s to workflow runs and depends on npm registry availability. Hosted runners now bake `@github/copilot` into toolcache (`$RUNNER_TOOL_CACHE/copilot-cli/<version>/<arch>/`). This change enables safe cached resolution while preserving existing fallback behavior.

## How it works

For Copilot engine workflows:

1. The compiler emits setup-step env vars **only for jobs that actually install/run Copilot CLI** (agent + threat-detection), and only in **release/action mode**:
   - `INPUT_INSTALL_COPILOT: 'true'`
   - `INPUT_GH_AW_VERSION=<compiler version>`
2. `setup.sh` checks `INPUT_INSTALL_COPILOT=true` and then invokes `actions/setup/js/install_copilot_cli.cjs` (guarded with `command -v node` and non-fatal behavior).
3. The resolver:
   - Fetches compat matrix from `gh-aw-actions/.github/aw/compat.json` (5s timeout via `AbortSignal.timeout`).
   - Falls back to bundled `actions/setup/compat.json` on error.
   - Selects compatible cached Copilot CLI from toolcache.
   - On hit: appends `<dir>/bin` to `$GITHUB_PATH`, sets `copilot-cached=true` and `copilot-path=<dir>`.
   - On miss/error: sets `copilot-cached=false` and exits 0.
4. The compiler-emitted installer step is gated with `if: steps.setup.outputs.copilot-cached != 'true'`, so it skips on cache hit and runs on miss.

## Diff shape

- `actions/setup/action.yml` — adds `copilot-cached`, `copilot-path` outputs.
- `actions/setup/setup.sh` — runs resolver only when explicitly opted in (`INPUT_INSTALL_COPILOT=true`).
- `actions/setup/compat.json` (new) — fallback matrix.
- `actions/setup/js/install_copilot_cli.cjs` (new) — resolver using native `fetch`.
- `actions/setup/js/install_copilot_cli.test.cjs` (new) — resolver unit tests.
- `pkg/workflow/copilot_engine_installation.go` — gates installer step on setup output.
- `pkg/workflow/compiler_yaml_step_generation.go` — emits Copilot resolver env only for opted-in jobs and only in release/action mode.
- `pkg/workflow/setup_step_version_test.go` — adds mode-sensitive coverage for env emission.
- Updated affected lock/golden outputs accordingly (including removing these env vars from dev-mode outputs).

## Backward compatibility

- Non-Copilot workflows: unchanged.
- Copilot workflows without compatible cache: installer runs as before.
- Runners without Node: resolver is skipped, installer runs as before.
- Dev/script-mode generated outputs no longer receive Copilot resolver env injection.

## Tests

- `make lint`
- `go test ./pkg/workflow -count=1`
- `make test`
- `make update-wasm-golden` (to refresh affected fixtures)
- JS resolver tests remain green (`install_copilot_cli.test.cjs`).

---

pr-sous-chef: branch update attempted for maintainer investigation. Run: https://github.com/github/gh-aw/actions/runs/26689893840

> Generated by [👨‍🍳 PR Sous Chef](https://github.com/github/gh-aw/actions/runs/26689893840) · gpt54 20.6M · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw%2Fpr-sous-chef%22&type=pullrequests)


actions/setup/js/install_copilot_cli.cjs
```diff
@@ -0,0 +1,309 @@
...
+  return new Promise(resolve => {
+    let settled = false;
+    const done = val => {
+      if (settled) return;
+      settled = true;
+      resolve(val);
+    };
+    let req;
+    try {
+      req = https.get(COMPAT_URL, { timeout: FETCH_TIMEOUT_MS }, res => {
```
pelikhan: Use fetch

pelikhan:
Just have it always on. It is more complicated to add a feature flag for this than anything.

actions/setup/js/install_copilot_cli.cjs
salmanmkc: Done in 3026304 — swapped `https.get` + buffer for `fetch` with `AbortSignal.timeout(5000)`. Matches `check_version_updates.cjs` / `send_otlp_span.cjs`. Same contract (returns parsed JSON or null on any error, never throws); 23/23 unit tests still pass.

actions/setup/js/install_copilot_cli.cjs
```diff
@@ -0,0 +1,268 @@
...
+//   { "max-gh-aw": "*"|<semver>, "min-agent": <semver>, "max-agent": <semver> }
+
+const fs = require("fs");
+const path = require("path");
+
+const COMPAT_URL = "https://raw.githubusercontent.com/github/gh-aw-actions/main/.github/aw/compat.json";
+const FETCH_TIMEOUT_MS = 5000;
+
+function log(msg) {
+  console.log(`[install_copilot_cli] ${msg}`);
```
pelikhan: You can require shim.cjs and use core.debug/info... for logging. Makes things more consistent for agents.

pelikhan (commented):
Add type annotations and typecheck

salmanmkc:
Done — dropped the feature flag entirely in fb6b0ef. The resolver now runs unconditionally for any Copilot-engine workflow (signal is the compiler-emitted `INPUT_GH_AW_VERSION` env). Non-Copilot workflows are unchanged (no env var, no resolver invocation, no golden diff). PR description updated to reflect the new shape.

pelikhan:
Actually... my bad this feature is only needed to run in agent/detection jobs + copilot engine selected.

actions/setup/js/install_copilot_cli.cjs
salmanmkc: Done in 8a55a3a — require shim.cjs at the top, swapped `console.log`/`console.error` for `core.info`/`core.warning`. Same `[install_copilot_cli]` prefix preserved for grep-ability.

salmanmkc:
Type annotations + typecheck addressed in 8a55a3a — added `// @ts-check` directive, JSDoc on every exported function, shared `ParsedSemver` and `CompatRow` typedefs, narrowed catch `unknown` to `Error` before reading `.message`. `npm run typecheck` is clean (the file was already in the tsconfig include glob; this just makes the contract explicit at the file level).

salmanmkc:
On it — bringing the flag back as compiler-emitted `INPUT_INSTALL_COPILOT: 'true'` on the setup step, gated in setup.sh on that env. Compiler only emits it for jobs that actually run Copilot CLI (the main agent job and the threat-detection job, when the Copilot engine is selected). Other consumers of the setup action (cache, activation, pre_activation, safe-outputs, unlock, publish-assets, repo-memory, notify-comment, experiments) will not trigger the resolver. Pushing in a moment.

salmanmkc:
Pushed `c806598`:

- Added `installCopilot bool` parameter to `generateSetupStep`; only `compiler_main_job.go` (agent) and `threat_detection.go` (detection) pass `true`. The other 9 callers pass `false`.
- When `installCopilot && setupEngineID == "copilot"`, the compiler emits both `INPUT_INSTALL_COPILOT: 'true'` and `INPUT_GH_AW_VERSION: "..."` on the setup step env block (script + dev/release modes).
- `setup.sh` resolver gate restored to `[ "${INPUT_INSTALL_COPILOT:-false}" = "true" ]`.
- Goldens regenerated: both env vars now appear only on agent jobs.
- Added `TestGenerateSetupStepEmitsInstallCopilotGate` (3 cases: copilot+opt-in, copilot+no-opt-in, non-copilot+opt-in). Full `pkg/workflow` test suite passes (260s).

Verified across 6 parallel review passes — no blockers; one acknowledged note that custom-command Copilot workflows (workflows that set `engine.command:`) will still receive the resolver opt-in but the resolver is a harmless no-op in that path since the custom script doesn't invoke the `copilot` binary. Happy to tighten further if you want.

pelikhan:
Did you consider updating the script that installs copilot instead?

salmanmkc:
> Did you consider updating the script that installs copilot instead?

Yeah, considered pure bash for the resolver during planning. The ADR specifically mandates @actions/tool-cache for .complete marker handling and semver-range matching, calling out that hand-rolled was the rejected alternative. Putting the lookup in install_copilot_cli.sh would be a similar hand-rolled path. Happy to revisit if the ADR's preference for @actions/tool-cache has shifted

.github/workflows/ace-editor.lock.yml
```diff
@@ -430,6 +430,8 @@ jobs:
           GH_AW_INFO_VERSION: "1.0.52"
           GH_AW_INFO_AWF_VERSION: "v0.25.56"
           GH_AW_INFO_ENGINE_ID: "copilot"
+          INPUT_INSTALL_COPILOT: 'true'
+          INPUT_GH_AW_VERSION: "947e72fa-dirty"
```
pelikhan: @copilot only inject this field in release mode

PR #37437, opened 2026-06-06
feat(max-ai-credits): allow -1 to disable budget enforcement and steering

dsyme:
### Summary

Introduces `-1` as a first-class **disable sentinel** for `max-ai-credits` / `default_max_ai_credits`. When set to `-1`, budget enforcement and token-steering are both suppressed end-to-end, giving operators a clean opt-out without having to remove the field entirely. All other negative values remain invalid.

Also includes a minor fix for pull-request finding (`91e194fc9`).

---

### Motivation

Previously the only way to run without a credit cap was to omit `max-ai-credits` altogether. This was fragile in environments where a default is set centrally (e.g. `GH_AW_DEFAULT_MAX_AI_CREDITS`) because there was no explicit override that meant "no limit." The `-1` sentinel provides a clear, documentable escape hatch that propagates correctly through every layer of the pipeline.

---

#### Schema & validation
| File | Change |
|---|---|
| `pkg/parser/schemas/main_workflow_schema.json` | Extended `max_ai_credits_limit` to accept `-1` as integer or string; updated field description. |
| `pkg/cli/env_command.go` | Switched `default_max_ai_credits` validation from `validatePositiveInt` → `validateNonZeroInt`. |

#### Runtime pipeline propagation
| File | Change |
|---|---|
| `pkg/workflow/engine_config_parser.go` | `parseMaxAICreditsValue` now delegates to the negative-pass-through normalizer instead of the positive-only one, so `-1` reaches `EngineConfig`. |
| `pkg/workflow/engine.go` | `ExtractEngineConfig` guard changed `> 0` → `!= 0` so `-1` is captured into the returned config. |
| `pkg/workflow/compiler_orchestrator_engine.go` | Both preserve-and-apply guards for `MaxAICredits` changed `> 0` → `!= 0` so the sentinel survives the compiler pipeline. |
| `pkg/workflow/compilerenv/manager.go` | `ResolveDefaultMaxAICredits` returns early with `-1` instead of falling back to the built-in default. |

#### Steering / budget enforcement
| File | Change |
|---|---|
| `pkg/workflow/awf_config.go` | Token-steering disable logic now triggers on negative `maxAICredits` in addition to negative `maxEffectiveTokens`; guard condition and log message updated accordingly. |

#### Documentation
| File | Change |
|---|---|
| `docs/src/content/docs/reference/cost-management.md` | Prose updated: `default_max_ai_credits` now described as accepting *non-zero integers*, with the `-1` disable semantic documented. |

---

### Tests added / updated

| File | What was changed |
|---|---|
| `pkg/parser/schema_test.go` | Two new functions: `max-ai-credits=-1` passes validation; any other negative (e.g. `-5`) is rejected. |
| `pkg/workflow/awf_config_test.go` | New case: `BuildAWFConfigJSON` omits both `enableTokenSteering` and `maxAiCredits` when `MaxAICredits == -1`. |
| `pkg/workflow/compilerenv/manager_test.go` | "Negative" test case updated: `GH_AW_DEFAULT_MAX_AI_CREDITS=-1` now expects `-1` returned (disable), not the fallback. |
| `pkg/cli/env_command_test.go` | Expected error message updated from "positive integer" → "non-zero integer". |

---

### Breaking changes

None. All existing positive `max-ai-credits` values behave identically. The only behaviour change is for the previously-invalid value `-1`.

---

### Migration

No action required for existing workflows. To explicitly disable budget enforcement in a workflow that inherits a central default:

```yaml
max-ai-credits: -1
```

Or via the environment:

```sh
gh aw env set GH_AW_DEFAULT_MAX_AI_CREDITS=-1
```

> Generated by [PR Description Updater](https://github.com/github/gh-aw/actions/runs/27076870614) for issue #37437 · 120.9 AIC · ⌖ 13.5 AIC · ⊞ 20.5K · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw%2Fpr-description-caveman%22&type=pullrequests)

### DIFF
--- a/pkg/parser/schemas/main_workflow_schema.json
+++ b/pkg/parser/schemas/main_workflow_schema.json
@@ -3709,7 +3709,7 @@
     "max-ai-credits": {
       "$ref": "#/$defs/max_ai_credits_limit",
       "default": 1000,
-      "description": "Per-run AI Credits budget control for firewall cost enforcement. Defaults to 1000 when omitted. Supports GitHub Actions expressions."
+      "description": "Per-run AI Credits budget control for firewall cost enforcement. Defaults to 1000 when omitted. Set to -1 to disable both budget enforcement and token steering. Supports GitHub Actions expressions."
     },
     "max-daily-effective-tokens": {
       "$ref": "#/$defs/max_daily_ai_credits_limit",
@@ -10554,8 +10554,20 @@
       ]
     },
     "max_ai_credits_limit": {
-      "description": "Maximum AI Credits budget for AWF API proxy enforcement. Accepts positive integers (including K/M suffix strings and expressions).",
-      "$ref": "#/$defs/positive_ai_credits_limit"
+      "description": "Maximum AI Credits budget for AWF API proxy enforcement. Accepts positive integers (including K/M suffix strings and expressions). Set to -1 to disable both budget enforcement and token steering.",
+      "oneOf": [
+        {
+          "$ref": "#/$defs/positive_ai_credits_limit"
+        },
+        {
+          "type": "integer",
+          "const": -1
+        },
+        {
+          "type": "string",
+          "pattern": "^-1$"
+        }
+      ]
     },
     "job_strategy": {
       "type": "object",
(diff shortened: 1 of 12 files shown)


pkg/parser/schemas/main_workflow_schema.json
```diff
@@ -10554,8 +10554,20 @@
       ]
     },
     "max_ai_credits_limit": {
```
pelikhan, line 10556: @dsyme looks like vibe junk. Will clean out the schema later.

pelikhan (approved):

pelikhan:
You need to create PR in draft so that "ready for review " triggers the reviewer AWs

PR #37887, opened 2026-06-08
Fix CI: add missing v0.3.24 MCPG pin and Windows test compatibility

dsyme:
Splits out the generic CI fixes from #37863 so they can land independently of that PR's behavior change.

### `test` job — `TestGetContainerPin_DefaultMCPImagesArePinned`
Adds the `ghcr.io/github/gh-aw-mcpg:v0.3.24` entry (with verified digest from GHCR) to `pkg/actionpins/data/action_pins.json`. The constant `DefaultMCPGatewayVersion` was bumped to `v0.3.24` in #37708 but the embedded pin entry was missed, so the default-image-is-pinned test has been failing on `main` since.

### `Windows Integration: compile` job (CWI workflow)
- `compilation_with_nonexistent_file`: relax the expected substring from `no such file` (POSIX-only) to `failed to read file`, which is what the engine itself emits on all platforms.
- `absolute_path_not_allowed`: pick an OS-appropriate absolute path so `filepath.IsAbs` returns true on Windows (`C:\\absolute\\path`).
- `TestCompileWithIncludeWithEmptyFrontmatterUnderPty`: skip on Windows where `creack/pty` returns 'unsupported'.

### `Windows Integration: version` job (CWI workflow)
- `TestVersionIsSetDuringBuild`: build the throwaway test binary inside `t.TempDir()` with the correct `.exe` suffix on Windows, instead of hardcoding `/tmp/gh-aw-test-version` which doesn't exist as an executable path on Windows runners.

## Verification

```
go test ./pkg/actionpins -run TestGetContainerPin          # PASS
go test -tags=integration -run 'TestCompileWorkflow$|TestCompileWorkflows_WorkflowDirValidation|TestCompileWithIncludeWithEmptyFrontmatterUnderPty' ./pkg/cli/...   # PASS
go test -tags=integration -run TestVersionIsSetDuringBuild ./cmd/gh-aw/...   # PASS
```

These same fixes are also present in #37863 and will continue to land there if this PR merges first.

### DIFF
--- a/cmd/gh-aw/version_test.go
+++ b/cmd/gh-aw/version_test.go
@@ -4,6 +4,8 @@ package main
 
 import (
 	"os/exec"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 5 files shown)


pelikhan:
Landon is bringing https://github.com/github/gh-aw/pull/37885

PR #37937, opened 2026-06-08
Stop retrying Copilot CLI PAT-rejection 400 as transient

pelikhan:
## Why

While investigating [run 27164602401](https://github.com/github/gh-aw/actions/runs/27164602401), the `Daily Ambient Context Optimizer` workflow failed with:

```
400 400 checking third-party user token: bad request: Personal Access Tokens are not supported for this endpoint
```

The supplied token (a PAT) is rejected by the Copilot CAPI endpoint, which is a persistent credential-type problem. The copilot-harness misclassified it as a generic partial-execution failure (`isAuthError=false isAuthenticationFailedError=false isCAPIError400=false`) and dutifully retried all 4 attempts with `--continue`, each producing the same 400 and wasting ~1 minute before failing.

## What

Extend `AUTHENTICATION_FAILED_PATTERN` in `actions/setup/js/copilot_harness.cjs` to also match the PAT-not-supported 400. This lets the existing first-attempt auth-failure short-circuit fire so the run fails fast with the real diagnostic instead of looping.

Added a unit test covering the new pattern branch.

## Note

This only changes the retry classifier. The underlying root cause (the workflow handing the Copilot CLI a PAT that the CAPI no longer accepts) still needs to be addressed separately by switching to a supported credential, e.g. a GitHub App installation token.


pelikhan:
@copilot apply copilot-review skill

pelikhan:
@copilot apply pr-finisher skill

Issue #40169, opened 2026-06-18
bug: call-workflow compiler propagates worker permissions to caller instead of checking them


dsyme:
## Summary

`test-copilot-call-workflow` has had `startup_failure` on every run since it was introduced. The immediate trigger is `vulnerability-alerts: read` appearing in the generated `call-test-copilot-call-worker` job's `permissions:` block, which GitHub Actions rejects. But the real bug is a design issue in `buildCallWorkflowJobs`: the compiler **propagates** the worker's permissions up into the caller's lockfile, when it should instead **check** that the caller's declared permissions are sufficient.

## Current (wrong) behaviour

`buildCallWorkflowJobs` in `compiler_safe_output_jobs.go` calls `extractCallWorkflowPermissions()` to read all job-level permissions from the worker's `.lock.yml` and unions them into a single set. That set is then written as the `permissions:` block on the `call-<worker>` job in the *caller's* lockfile (comment at line 171: "includes a job-level permissions block that is the union of all the worker's job-level permissions, so GitHub allows the nested jobs to run").

This is the wrong approach: the caller's compiled permissions block should be derived from the caller's own frontmatter/declared permissions, not reverse-engineered from the worker's implementation. The worker's permissions are the worker's business; the caller should declare its own scope and be told if that's insufficient.

## How the `vulnerability-alerts` symptom appears

The worker's `agent` job has `permissions: read-all`. When `extractJobPermissionsFromParsedWorkflow` merges that shorthand with the explicit-scope maps from other jobs, `Merge()` in `permissions_operations.go` expands `read-all` by iterating over `GetAllPermissionScopes()`. That function includes `PermissionVulnerabilityAlerts = "vulnerability-alerts"` (declared as a scope GitHub is rolling out). The expansion materialises `vulnerability-alerts: read` as an explicit key, which gets written into the caller's lockfile. GitHub Actions rejects the workflow at parse time: `startup_failure`.

The actionlint warning for this scope is deliberately suppressed in `lint_command.go` with the comment "GitHub is rolling out an additional permissions scope before actionlint support" — but real runs confirm GitHub does **not** accept it for `GITHUB_TOKEN`, so the suppression masks a genuine error.

## Proposed fix

[...]

PR #40175, opened 2026-06-19
fix: derive call-workflow job permissions from caller, not worker (#40169)

dsyme:
## Summary

Fixes a bug where `buildCallWorkflowJobs` was deriving the permission envelope of a generated call-workflow job from the *worker's* resolved permissions rather than the *caller's* declared permissions. This could cause generated workflows to run with an inflated or incorrect permission set that did not match what the caller author intended.

## Problem

When a `call-workflow` job is compiled, the compiler previously set `callJob.Permissions` to the union of the worker's job-level permissions (`rendered`). The worker's permissions are derived dynamically and represent what the worker *needs*, not what the caller has explicitly declared. This violated the principle that a caller should never be granted permissions it did not ask for.

### `pkg/workflow/compiler_safe_output_jobs.go` *(high impact)*
- **Removed** the unconditional `callJob.Permissions = rendered` assignment that propagated worker permissions onto the generated job.
- **Added** two new outcomes in its place:
  1. A **non-fatal skip** when the worker file is not yet resolved (e.g., in a batch where order matters).
  2. A **missing-permission warning** (non-fatal) surfaced via `formatCompilerMessage` + `os.Stderr` + `c.IncrementWarningCount()` when the caller's declared permissions do not fully cover the scopes the worker requires.
- The call-worker job's permission envelope is now sourced exclusively from the **caller's own declared `Permissions` field**.

### `pkg/workflow/call_workflow_permissions.go` *(high impact)*
- Added `permissionLevelRank` helper to compare permission levels by severity order.
- Added `findUncoveredWorkerPermissions` helper that returns the set of worker-required scopes not covered (at sufficient level) by the caller's declared permissions.
- Fixed a missing `worker.Get(scope)` call inside `findUncoveredWorkerPermissions` (commit `e34e404c1`).
- Clarified `extractCallWorkflowPermissions` doc comment: the function is used **only for validation**, never to inflate caller permissions.

### `pkg/workflow/call_workflow_permissions_test.go` *(medium impact)*
- Added `TestFindUncoveredWorkerPermissions` with six sub-cases covering:
  - fully-covered permissions (expect no warnings)
  - partially-covered and uncovered scopes
  - level-rank boundary cases (e.g., `read` vs `write`)
- Revised existing end-to-end and YAML-output tests to assert that generated call-workflow jobs carry the **caller's** declared permissions, not the worker's.

### `pkg/cli/workflows/test-copilot-call-workflow.md` *(low impact)*
- Added `issues: read` and `pull-requests: read` to the test workflow's top-level `permissions` block to align the fixture with the new validation logic.

## Behaviour Change

| Scenario | Before | After |
|---|---|---|
| Caller declares `contents: read`; worker needs `contents: write` | Job silently gets `contents: write` | Job gets `contents: read`; compiler emits a warning that `contents: write` is uncovered |
| Caller declares all required scopes | Job gets worker-derived permissions | Job gets caller's declared permissions (same result in practice) |
| Worker file not yet resolved in batch | Crash / wrong permissions | Non-fatal skip; job is not emitted for this compile pass |

## Risk & Compatibility

- **Breaking:** No — no public API, CLI flag, or config key is renamed or removed.
- **Compiler output:** Permission envelopes on generated call-workflow jobs will now differ from previous output when the caller did not explicitly declare the worker's required scopes. This is a correctness fix.
- **Warnings:** Projects where callers under-declare permissions will now see compiler warnings. These are non-fatal; the workflow is still emitted.

> Generated by [PR Description Updater](https://github.com/github/gh-aw/actions/runs/27902188425) for issue #40175 · 119.3 AIC · ⌖ 7.6 AIC · ⊞ 4.5K · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw%2Fpr-description-caveman%22&type=pullrequests)


dsyme:
@copilot fix the CI errors

PR #41476, opened 2026-06-25
feat: Introduce Impact Score feature

mnkiefer:
- Introduces **Impact Score**, which assigns each issue or PR a deterministic repo-defined priority score from mapping `aw.json` to help identify the highest-impact work, which agentic workflows produced it, and whether that value is worth the observed cost.
  - Impact Score builds a graph of GitHub issues, PRs, labels, workflows, and signals, extracts item features from that graph, applies the repo’s `aw.json` scoring rules, clamps each score, and ranks work items and workflows by the resulting impact:
  
      `impact_score = clamp(base + effects from matching aw.json rules)`
  - Policy language is just safe JSON, not code, with simple rules of:
    - **when**: conditions that match issue/PR facts
    - **score**: set the score
    - **min**: raise score to at least a value
    - **max**: cap score before final clamp
    - **add**: add points
    - **stop**: stop after this rule matches


mnkiefer:
@pelikhan This is `v0.1.0` of the Impact Scoring. Happy to talk about the (graph) model, initial bootstrap & default values for impact computation. 👍🏼  Since we can't use a custom function (to be able to cover any user's impact value, no matter how complex)due to security concerns, we now  rely on the JSON policy language which is restricted and we must expand on demand.

> How can we compute the impact score in the action runtime to display it in the generated footer?

In action footers, the workflow can compute the score by reading aw.json, normalizing the current issue or PR fields, and running the same local scoring rules.

> How can we maintain the mapping to prevent rot?

To prevent rot, the mapping should use GitHub-native fields where possible, keep derived fields centralized, and warn when rules reference labels, workflows, or dimensions that no longer match.

mnkiefer:
- Moving to a separate repository for evals

PR #50156, opened 2026-08-04
Fix workflow update source-only merges

pelikhan:
## Summary
- exclude the updater-managed `source` field from three-way merge inputs
- fast-path source-only upstream changes to preserve local workflow formatting and avoid false conflict markers
- update `ci-doctor` to `42c2ab5` and add regression coverage

## Validation
- `go test ./pkg/cli -run TestMergeWorkflowContent`
- `./gh-aw compile ci-doctor`
- `make agent-report-progress`


pelikhan:

PR #50929, opened 2026-08-06
fix(sanitizer): align URL authority parsing with browser URL parsing

pelikhan:
### Overview

Fixes a family of URL-allowlist bypasses in the content sanitizer (`actions/setup/js/sanitize_content_core.cjs`) caused by disagreement between the sanitizer's regex-based URL parsing and how browsers/URL parsers actually resolve authority (userinfo/host). An attacker could craft a URL that displays an allowlisted host to the sanitizer while a browser resolves a different, attacker-controlled host — enabling a zero-click exfiltration channel via GitHub's camo image proxy.

### Vulnerability class

Everything before the last `@` in a URL authority is userinfo (credentials), not the host. `(evil.com/redacted) and `(evil.com/redacted) both connect to `evil.com`, even though `github.com` appears first and could pass a naive allowlist check. Embedded in a markdown image (`x, this becomes a zero-click exfiltration vector since GitHub's camo proxy fetches image URLs server-side when a comment is rendered.

### Key changes

- **`actions/setup/js/sanitize_content_core.cjs`**
  - Added `URL_START_DELIMITERS` and `URL_AUTHORITY_CHAR` shared regex components so the userinfo-stripping pass and the domain-filtering pass use identical boundary rules and cannot disagree.
  - Added `stripUrlIgnorableWhitespace()` to discard tab/CR/LF inside an authority before host comparison, matching browser URL-parser preprocessing (previously these were treated as terminators, allowing bypass).
  - Added `stripProtocolRelativeUserinfo()` to handle userinfo spoofing in protocol-relative URLs (`(evil.com/redacted)`), which `stripUrlUserinfo()` alone could not cover since it requires a scheme.
  - Extended the URL-start delimiter set to include `<` and `=` so HTML attributes and CommonMark angle-bracket destinations are examined.
  - Recognized and normalized backslash separators (`\host`, `(host/redacted)`), since URL parsers treat `\` as `/`.
  - Fixed authority parsing so it no longer swallows the delimiter of a following URL (e.g. `(a.com/redacted),(evil.com/redacted) previously let the second URL escape stripping entirely).

- **`actions/setup/js/sanitize_content.test.cjs`**
  - Added a dedicated "URL userinfo authority spoofing" test suite (25+ cases) covering `https://`/protocol-relative URLs, markdown images, HTML attributes, backslash separators, and whitespace-based bypass attempts.

- **`.changeset/fix-protocol-relative-url-userinfo-bypass.md`**
  - Added patch changeset describing the fix and the four closed differentials.

### Security impact

- **Severity**: High — allowlist bypass in a security-relevant sanitizer, exploitable as a zero-click data-exfiltration channel through GitHub's camo proxy.
- **Breaking change**: No. Purely tightens sanitization; no public API changes.

### Testing

- New unit tests in `sanitize_content.test.cjs` directly exercise each bypass form (userinfo spoofing, protocol-relative, backslash separators, embedded whitespace) and assert the spoofed host is redacted rather than the allowlisted one.> Generated by [PR Description Updater](https://github.com/github/gh-aw/actions/runs/31135528489) for #50929 · auto · 44 AIC · ⊞ 6.8K · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw%2Fpr-description-caveman%22&type=pullrequests)


pelikhan:

PR #55155, opened 2026-08-23
feat: add value grader

mnkiefer:
- Adds an experimental, opt-in value grader that measures whether each workflow run achieved its intended repository outcome.
- Unlike existing graders, it measures operational value rather than execution quality.
```yaml
graders:
  operational-value:
    run: .github/graders/WORKFLOW-NAME-operational-value.sh
```
-  Configured workflows automatically record value in [0,1], evidence provenance, maturity, and optional baseline delta.
- Runs ca be regrdaded as evidence matures:
`gh aw graders value RUN-ID --evidence-at TIMESTAMP --json`


docs/src/content/docs/reference/trace-graders.md
```diff
@@ -56,12 +56,39 @@ graders:
...
 Custom scripts must return a value and stay within 4096 characters (no `require`, `import`, `fetch`, `eval`, or `process.exit`).
 
+## Operational value grader
+
+Configure the reserved `value` grader with a repository-relative Bash function:
+
+```aw wrap
+graders:
+  value:
+    function: .github/graders/daily-file-diet-value.sh
```
pelikhan: Use "run:" to match typical naming convention for shell scripts

pelikhan:
"value" and "function" are generic terms, we should try to find more esoteric/unique terms to describe this feature to help downstream with search/grep

mnkiefer:
> "value" and "function" are generic terms, we should try to find more esoteric/unique terms to describe this feature to help downstream with search/grep

I agree, renamed back to "operational value" (as in my original POC) and "evaluator" now.

pelikhan:
The fact that it widens the github actions token is somewhat a problem. We might need to move this to another job.

mnkiefer:
> The fact that it widens the github actions token is somewhat a problem. We might need to move this to another job.

I moved the workflow-run lookup to the activation job, which already has compiler-controlled permissions. It passes `created_at` to the grader as an output, so the main agent token is no longer widened with `actions: read`.

mnkiefer:
@copilot Fix PR related tests

Issue #59268, opened 2026-09-07
Isolate the enclave delegation control listener from executor networks


lpcox:
## Summary

Make mcpg's dynamic-enclave delegation control listener unreachable from the primary agent and single-use enclave executor networks.

The controller bootstrap added by #59045 / PR #59046 publishes the control port on host loopback and documents that this prevents the enclave executor network from routing to it. That claim is incorrect when mcpg listens on `0.0.0.0` inside its container: Docker's `-p 127.0.0.1:<host>:<container>` limits access through the host-published port, but a peer sharing a Docker network with mcpg can address the container IP and control port directly without traversing host loopback.

The control API still requires the AWF-only 256-bit capability, so this is not an unauthenticated authorization bypass. It is a missing defense-in-depth/topology guarantee and conflicts with `github/gh-aw-firewall#8195`, which requires the dynamic executor to stay off the control listener.

## Scope

- Separate the delegation control transport from the mcpg data-plane transport at the container/network boundary.
- Ensure the primary agent, enclave executor, model sidecar, and general MCP data-plane peers have no route to the control listener.
- Preserve host-side AWF access to the control API and keep the existing capability authentication on every request.
- Do not rely solely on `-p 127.0.0.1:...` publication scope when the same in-container listener is reachable through a shared bridge network.
- Correct the inaccurate reachability comment in `pkg/workflow/enclaves.go` and related documentation/tests.
- Add a topology test that attempts a TCP connection to the control port from the primary and enclave executor network namespaces and proves both fail, while an authenticated AWF host request succeeds.
- Preserve the executor-facing mcpg data plane and its invocation-scoped bearer behavior.
- Fail closed if the isolated control transport cannot be established; do not fall back to a shared listener reachable by executors.

## Possible implementations

- Run the control plane in a separate sidecar/network namespace reachable only through an AWF-owned Unix socket or host-private network.
- Bind the control listener to an interface/address not attached to the executor-visible bridge while keeping the data listener on the existing mcpg container network.
- Use a host-owned authenticated relay with no executor-visible listener.

The implementation must demonstrate network separation rather than assuming host publication rules apply to container-to-container traffic.

## Acceptance criteria

- From the enclave executor network namespace, connecting directly to mcpg's control port fails at the network layer.
- From the primary agent network namespace, connecting to the control port also fails.
- AWF can still perform authenticated create/status/reconcile/revoke operations.
- Possession of an executor bearer does not provide a route or credential for the control API.
[...]

PR #59310, opened 2026-09-07
docs(enclaves): correct delegation control listener reachability claim

lpcox:
## Summary

Corrects a factually incorrect security comment on mcpg's delegation control listener, and records the accepted design: **capability authentication is the enforced control on that listener**, not network topology.

Comment-only change. No behavior change, no generated-output change.

## The incorrect claim

`buildMCPGatewayContainerCommand` publishes the control port with `-p 127.0.0.1:<port>:<port>` and asserted:

> the primary-agent network, enclave executor network, and any docker-sbx guest **cannot route to it** — only the AWF host process reaches it

That conclusion doesn't follow from its own premise. `-p 127.0.0.1:X:X` constrains the **host-published** path only. The in-container listener is deliberately bound to `0.0.0.0` in bridge mode (`writeMCPGatewayExports`, and the comment there says why — Docker NATs a published port to the container's bridge IP, so a container-local `127.0.0.1` bind would be unreachable). A peer co-attached to any network mcpg joins reaches the container IP directly and never traverses the published port.

This isn't hypothetical. AWF asserts the co-attachment on every dynamic run — mcpg is a required member of the enclave-agent network at `172.31.0.40`, which is how the enclave reaches the GitHub data plane at all:

```js
// gh-aw-firewall containers/enclave/agent-executor/docker-enclave-runner.js:35-39
if (this.config.githubEnabled || this.config.dynamicEnabled) {
  const steadyStateMembers = [proxyMember,
    `${this.config.githubGatewayContainer}@172.31.0.40/24,`];
```

Control sits on the same interface at data-port + 10 (`enclaveDelegationControlPortOffset = 10`).

## Why this is a comment fix and not a topology change

Per maintainer decision, we accept the 256-bit capability as the control here rather than adding a second, topological one. That's defensible on the implementation as it stands:

| Property | Where |
|---|---|
| 256-bit CSPRNG secret, `::add-mask::`-ed | `mcp_setup_gateway.go:315-316` |
| Excluded from the agent environment via `--exclude-env` | `awf_env.go:108`, `:174` — and the exclusion condition is *broader* than the setting condition |
| Stored only as a SHA-256 digest, never in comparable form | mcpg `internal/delegation/capability.go` |
| `subtle.ConstantTimeCompare` over fixed-size digests | same — doesn't leak length or content |
| 32-byte minimum enforced at construction | `minControlCapabilityBytes` |
| Verified **before any control routing**, plus a POST requirement | mcpg `internal/proxy/delegation.go:49` |
| Served by a handler separate from the executor data plane | `ControlHandler()` vs `Handler()` — a valid executor bearer cannot reach control ops |

So the residual risk is capability *leakage*, not brute force. Network isolation would only have added depth against that. The comment now says this plainly instead of implying a routing guarantee that doesn't hold.

The `writeMCPGatewayExports` comment was accurate ("never reachable from outside the host") but misleading by omission, so it gets a pointer rather than a rewrite.

## Validation

- `gofmt -l` clean
- `go build ./pkg/workflow/` clean
- `go vet ./pkg/workflow/` clean
- `go test ./pkg/workflow/ -run 'Enclave|Delegation|AWFExclude'` — pass

Existing coverage already locks the properties we're relying on (`enclaves_test.go:671-680`): capability minting, masking, `GITHUB_ENV` handoff, and exclusion of both the capability and the control endpoint from the agent environment. No new test is needed for a comment change, and I confirmed the exclusion assertions exist rather than adding a duplicate.

Refs github/gh-aw#59268
Refs github/gh-aw-firewall#8195

---

> Generated by [👨‍🍳 PR Sous Chef](https://github.com/github/gh-aw/actions/runs/34168294676) · pi · gpt54 · 40.3 AIC · ⌖ 8.65 AIC · ⊞ 9.2K · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw%2Fpr-sous-chef%22&type=pullrequests)
> <sub>Comment <em>/souschef</em> to run again</sub>

### DIFF
--- a/pkg/workflow/enclaves.go
+++ b/pkg/workflow/enclaves.go
@@ -53,13 +53,14 @@ const (
 	maxDynamicEnclaveTimeoutSeconds = 4740
 	enclaveMCPTransportAllowance    = 60
 	// enclaveDelegationControlPortOffset is added to the job's MCP gateway data-plane
-	// port to derive the private, host-only listener port for mcpg's
-	// github-repository-delegation-v1 control plane. Deriving it from the (per-job
-	// configurable) data-plane port, rather than a single fixed literal, avoids a
-	// port collision on self-hosted runners that execute multiple concurrent jobs
-	// with distinct gateway ports on the same host. The control port is bound
-	// separately from MCP_GATEWAY_PORT and published only to loopback so neither
-	// the primary agent nor the enclave executor network can route to it.
+	// port to derive the listener port for mcpg's github-repository-delegation-v1
+	// control plane. Deriving it from the (per-job configurable) data-plane port,
+	// rather than a single fixed literal, avoids a port collision on self-hosted
+	// runners that execute multiple concurrent jobs with distinct gateway ports on
+	// the same host. The control port is bound separately from MCP_GATEWAY_PORT and
+	// published only to host loopback; in bridge mode, co-attached containers can
+	// still reach the in-container listener directly, so capability authentication
+	// is the enforced control.
 	//
 	// Contract: this offset only avoids collisions between a single job's own
 	// data-plane and control ports. Operators running self-hosted runners with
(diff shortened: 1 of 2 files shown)


lpcox:
@copilot  address the review feedback

Issue #263, opened 2025-08-31
Test aliases


dsyme:

dsyme:
@test-claude-alias What is 1+1

---
*🤖 [Workflow run](https://github.com/githubnext/gh-aw/actions/runs/17363818125) triggered by this comment*

PR #60682, opened 2026-09-13
Simplify operational-value grading to one-shot evaluators

mnkiefer:
## Summary

- Replace historical replay, maturation, baseline, cache, registry, and report machinery with a deterministic one-shot Bash evaluator contract.
- Freeze inline or file-backed evaluator bytes and their SHA-256 digest into each compiled workflow run, with strict metric-array validation and size limits.
- Supply current-run validated safe-output requests as `request.outputs` while preserving the distinction between requested and applied mutations.
- Refocus the operational-value designer on ultimate goals and the strongest observable, attributable, independently verifiable outcome; remove domain-specific metric templates.
- Add a Daily File Diet evaluator that independently checks the largest non-test Go file at the run SHA against the workflow's 800-line threshold and validates the required issue/noop decision.

## Why

Operational value should measure the strongest effect a run can actually prove at grading time. Workflow-specific semantics now live in frozen evaluators instead of generic historical infrastructure, avoiding hindsight tuning and claims about future or unapplied outcomes.

## Validation

- Operational-value designer tests and deterministic evaluator fixtures pass.
- Bash syntax and prospective contract-change verification pass.
- Historical Daily File Diet run #263 evaluates to `file-diet-decision-conformance=1` and `largest-file-under-threshold=0` under the new evaluator.
- All 299 workflows recompile successfully with lock files in sync.
- `make agent-report-progress` passes.


.github/skills/operational-value-designer/SKILL.md
```diff
@@ -71,80 +206,56 @@ The evaluator uses Bash 3.2-compatible Bash plus `jq` and supports:
...
-Add `--repo [HOST/]OWNER/REPO` to select the GitHub host for the current repository checkout. The command downloads the original grader artifact, reuses its operational case and complete run subject, and refuses to execute unless the archived evaluator matches both digest records and the evaluator at the recorded commit in the trusted checkout. It prints a new observation and never modifies the original artifact.
+### 8. Verify and review
 
-## Build a Historical Report
-
-Build a report across all completed runs from the contract's adoption time:
+Run:
 
 ```bash
-gh aw graders operational-value report WORKFLOW-NAME
```
pelikhan, line 119: The cli can run graders. We should also support embedded graders without .sh file

